PR #11, opened 2022-06-11
Batch & Session APIs

rathorevaibhav:
### Summary
More API endpoints for:
1. Batch bulk update sessions
2. Session bulk update batches
3. Correction in batch session migration for batch instead of the user. Will have to refresh migrations and recreate DB.
4. Hyphenated URLs

### DIFF
--- a/lib/dbservice/batches.ex
+++ b/lib/dbservice/batches.ex
@@ -118,4 +118,21 @@ defmodule Dbservice.Batches do
     |> Batch.changeset_update_users(users)
     |> Repo.update()
   end
+
+  @doc """
+  Updates the sessions mapped to a batch.
+  """
+  def update_sessions(batch_id, session_ids) when is_list(session_ids) do
+    batch = get_batch!(batch_id)
+
+    sessions =
+      Dbservice.Sessions.Session
+      |> where([session], session.id in ^session_ids)
+      |> Repo.all()
+
+    batch
+    |> Repo.preload(:sessions)
+    |> Batch.changeset_update_sessions(sessions)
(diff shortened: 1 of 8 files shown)


lib/dbservice/sessions.ex
```diff
@@ -102,6 +102,23 @@ defmodule Dbservice.Sessions do
     Session.changeset(session, attrs)
   end
 
+  @doc """
+  Updates the users mapped to a batch.
+  """
+  def update_batches(session_id, batch_ids) when is_list(batch_ids) do
```
pritamps, line 108: There's no user anywhere in this method, though users are mentioned in the doc. This is intentional?

lib/dbservice/batches.ex
```diff
@@ -118,4 +118,21 @@ defmodule Dbservice.Batches do
     |> Batch.changeset_update_users(users)
     |> Repo.update()
   end
+
+  @doc """
+  Updates the sessions mapped to a batch.
+  """
+  def update_sessions(batch_id, session_ids) when is_list(session_ids) do
```
pritamps, line 125: The Swagger PR has API docs, no? I feel like it will be needed :D

lib/dbservice/sessions.ex
rathorevaibhav: Sorry copy/paste error. I'll update.

lib/dbservice/batches.ex
rathorevaibhav: Yes, coming after a couple of PRs.

pritamps (approved):

PR #12, opened 2022-06-13
#10 User session mapping with additional data

rathorevaibhav:
Targets #10 

1. User session mapping
2. User session endpoints for CRUD operations
3. Sending users for a session occurrence along with session occurrence data - only for single session occurrence API call


pritamps (approved):
This looks good to me, but I think maybe it's time for a meeting to explain how the APIs will work and all. I think I've lost track a bit!

rathorevaibhav:
Sure. Let's catchup early next week @pritamps I'll ping over Discord

PR #16, opened 2022-06-20
Seeding fake data

rathorevaibhav:
Targets #19 

### Summary
At times, we may want to populate the data with some fake data on a fresh DB (maybe at the time of project setup) so that we can start experimenting with API changes quickly. This PR adds a one-liner command to randomly seed DB with fake data.

1. Added elixirs/faker library as mix dependency
2. Modified UUID column in session table to support Elixir/postgres default UUID generation
3. Faker initialized for unit testing
4. Documentation update for seeding data
5. `priv/repo/seeds.exs` updated to add seed data for all the entities


sudheshna-donthineni (approved):
It looks fine to me!

PR #18, opened 2022-06-28
Response data updates

rathorevaibhav:
### Summary
1. Added some missing fields in the response data for session, enrollment record, student and teacher
2. Added some missing fields in the request data (cast) for session and session occurrence.


sudheshna-donthineni (approved):
I assume these fields have been discussed with @pritamps or Karn. Everything else looks fine!

rathorevaibhav:
Yes @sudheshna-donthineni 
These fields are in schema but I missed them while putting in the API payload/response. These do not get auto-included when I ran the command to create controllers, migrations etc.

PR #25, opened 2022-07-19
#8 Single API for student & user registration

rathorevaibhav:
Requires #16 to be merged before

Targets #8 

### Summary
1. Single endpoint where the API client can send both user and student details in the payload and the service will create a user and a student for that.
2. The errors for user and student validation will happen as it is
3. Fixes in `belongs_to` for several structs as I figured that we need to explicitly import all the dependencies in Elixir to make things work as expected
4. `show_with_user.json` for student response to also include associated user info


pritamps (approved):
Looks good to me!'

PR #32, opened 2022-09-12
Latest schema updates!

Bahugunajii:
### Description
- Added new tables and modified existing ones
- Made all schema changes
- Removed batches and all other files associated to it
- Made seeders for newly created tables

### Checklist
- [x] Performed self-review of my own code


lib/dbservice/groups/group.ex
```diff
@@ -4,18 +4,56 @@ defmodule Dbservice.Groups.Group do
...
+    field :program_product_used, :string
+    field :program_donor, :string
+    field :program_state, :string
+    field :batch_contact_hours_per_week, :integer
+    field :group_input_schema, :map
+    field :group_locale, :string
+    field :group_locale_data, :map
+
+    has_many :student, Student
+    has_many :group_student, GroupStudent
```
rathorevaibhav, line 30: let me check a bit on this. I think we can skip the connection between a group and a student and rather use group user connection only. We can anyway find the student related to the user, which in turn would be related to the group.

Let's remove group_student altogether

lib/dbservice/users/student.ex
```diff
@@ -15,8 +16,20 @@ defmodule Dbservice.Users.Student do
...
+    field :father_education_level, :string
+    field :mother_profession, :string
+    field :mother_education_level, :string
+    field :time_of_device_availability, :date
+    field :has_internet_access, :boolean
+    field :primary_smartphone_owner, :string
+    field :primary_smartphone_owner_profession, :string
     belongs_to :user, User
     belongs_to :group, Group
+    has_one :enrollment_record, EnrollmentRecord
```
rathorevaibhav, line 32: a student can have more than one enrollment records. An enrollment record represents an academic year. Hence, as a student gets into the next year, there'll be a new enrollment record. The old one will be marked as inactive and the new one will be active.

lib/dbservice/groups/group.ex
```diff
@@ -4,18 +4,54 @@ defmodule Dbservice.Groups.Group do
...
+  alias Dbservice.Users.User
+  alias Dbservice.Sessions.Session
+
   schema "group" do
-    field :input_schema, :map
-    field :locale, :string
-    field :locale_data, :map
+    field :name, :string
+    field :parent_id, :integer
+    field :type, Ecto.Enum, values: [:batch, :group, :cohort, :program]
```
rathorevaibhav, line 14: I've generally seen in applications that when we try to modify an enum value, it gives a lot of trouble. Can you check once the following scenario:
1. Run all the migrations
2. Put in some data in this table with some dummy values
3. Try adding another value to this enum field
4. Generate and run the migration

If these steps are running fine without any errors, then we're all good. Maybe let's create an issue for this and you can work on it separately.

lib/dbservice/users/student.ex
```diff
@@ -15,8 +16,20 @@ defmodule Dbservice.Users.Student do
     field :mother_phone, :string
     field :stream, :string
     field :uuid, :string
+    field :physically_handicapped, :boolean
+    field :cohort, :string
```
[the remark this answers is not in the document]
rathorevaibhav: Did you confirm this? Any update?

lib/dbservice/users/student.ex
```diff
@@ -15,13 +16,32 @@ defmodule Dbservice.Users.Student do
...
 
   @doc false
+  @spec changeset(
+          {map, map}
+          | %{
+              :__struct__ => atom | %{:__changeset__ => map, optional(any) => any},
+              optional(atom) => any
+            },
+          :invalid | %{optional(:__struct__) => none, optional(atom | binary) => any}
+        ) :: Ecto.Changeset.t()
```
rathorevaibhav, line 44: I think we had not been adding spec for changesets. That's why `@doc false` was there I believe. Can you share how this is showing up on swagger?

priv/repo/seeds.exs
```diff
@@ -40,27 +41,31 @@ defmodule Seed do
...
+        type: Enum.random(["batch", "group", "cohort", "program"]),
+        program_type: Enum.random(["Competitive", "Non Competitive"]),
+        program_sub_type: Enum.random(["Easy", "Moderate", "High"]),
+        program_mode: Enum.random(["Online", "Offline"]),
+        program_start_date: Faker.DateTime.backward(Enum.random(1..10)),
+        program_target_outreach: Enum.random(["1", "2", "3", "4", "5", "6", "7", "8", "9", "10"]),
+        program_product_used: Enum.random(["One", "Less than 5", "More than 5"]),
+        program_donor: Enum.random(["YES", "NO"]),
+        program_state: Enum.random(["UK", "UP", "MP", "HP", "PB"]),
+        batch_contact_hours_per_week: Enum.random(["32", "40"]),
```
rathorevaibhav, line 65: isn't this an integer field in db? So we should insert int values?

priv/repo/seeds.exs
```diff
@@ -40,27 +41,31 @@ defmodule Seed do
...
-        input_schema: %{},
-        locale: Enum.random(["hi", "en"]),
-        locale_data: %{}
+        name: Person.name(),
+        type: Enum.random(["batch", "group", "cohort", "program"]),
+        program_type: Enum.random(["Competitive", "Non Competitive"]),
+        program_sub_type: Enum.random(["Easy", "Moderate", "High"]),
+        program_mode: Enum.random(["Online", "Offline"]),
+        program_start_date: Faker.DateTime.backward(Enum.random(1..10)),
+        program_target_outreach: Enum.random(["1", "2", "3", "4", "5", "6", "7", "8", "9", "10"]),
```
rathorevaibhav, line 61: let's make this a values in thousands. Like how many thousand students the program is targeting. 1 to 10 is way less

priv/repo/seeds.exs
```diff
@@ -143,7 +150,18 @@ defmodule Seed do
         category: Enum.random(["General", "OBC", "SC", "ST"]),
         stream: Enum.random(["Science", "Commerce", "Arts"]),
         user_id: user.id,
-        group_id: group.id
+        group_id: group.id,
+        physically_handicapped: Enum.random([true, false]),
+        course: Enum.random(["JEE", "NEET", "NDA"]),
+        family_income: Enum.random(["1LPA-3LPA", "3LPA-6LPA", ">6LPA"]),
+        father_profession: Enum.random(["Labour", "Private", "Government"]),
```
rathorevaibhav, line 157:
```suggestion
        father_profession: Enum.random(["Private", "Government", "Self-employed", "Not-employed"]),
```

priv/repo/seeds.exs
```diff
@@ -143,7 +150,18 @@ defmodule Seed do
...
+        father_education_level: Enum.random(["UG", "PG", "NA"]),
+        mother_profession: Enum.random(["Housewife", "Private", "Government"]),
```
rathorevaibhav, line 159:
```suggestion
        mother_profession: Enum.random(["Housewife", "Private", "Government", "Self-employed", "Not-employed"]),
```

rathorevaibhav (approved):

PR #45, opened 2022-11-14
API to create & update student with a user and teacher with a user

Bahugunajii:
### Description
- Created an API which creates & updates a student with a user
- Created an API which creates & updates a teacher with a user

### Checklist
- [x] Performed self-review of my own code

### DIFF
--- a/lib/dbservice/users.ex
+++ b/lib/dbservice/users.ex
@@ -225,6 +225,28 @@ defmodule Dbservice.Users do
   end
 
   @doc """
+  Updates a user first and then the student.
+
+  ## Examples
+
+      iex> update_student_with_user(%{field: value})
+      {:ok, %Student{}}
+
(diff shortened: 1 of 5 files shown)


lib/dbservice/users.ex
```diff
@@ -224,6 +224,28 @@ defmodule Dbservice.Users do
...
+  ## Examples
+
+      iex> update_student_with_user(%{field: value})
+      {:ok, %Student{}}
+
+      iex> update_student_with_user(%{field: bad_value})
+      {:error, %Ecto.Changeset{}}
+
+  """
+  def update_student_with_user(student, users, attrs \\ %{}) do
```
rathorevaibhav, line 239: is it supposed to be `users` or `user`? As I understand, a student is only connected to one user, right?

rathorevaibhav (approved):

PR #47, opened 2022-11-22
Added filter query!

Bahugunajii:
## Description
- Added query to filter results
- Modified migrations

## Output response
image

## Checklist
- [x] Self review

### DIFF
--- a/lib/dbservice_web/controllers/session_controller.ex
+++ b/lib/dbservice_web/controllers/session_controller.ex
@@ -1,6 +1,8 @@
 defmodule DbserviceWeb.SessionController do
   use DbserviceWeb, :controller
 
+  import Ecto.Query
+  alias Dbservice.Repo
   alias Dbservice.Sessions
   alias Dbservice.Sessions.Session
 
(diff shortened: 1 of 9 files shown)


lib/dbservice_web/controllers/session_controller.ex
```diff
@@ -27,8 +29,23 @@ defmodule DbserviceWeb.SessionController do
...
+
+        {"portal_link", portal_link}, query ->
+          from(u in query, where: u.portal_link == ^portal_link)
+
+        {"session_id", session_id}, query ->
+          from(u in query, where: u.session_id == ^session_id)
+
+        _, query ->
+          query
+      end)
```
rathorevaibhav, line 46: Can't we do something like this? I've written pseudo code but hopefully it's straightforward and easy to understand
```
if (params) {
   foreach (param in params) {
     query = {param, param_value}, query -> from(u in query, where u[param] = param_value)
   }
}
```

So the idea is not hardcode any columns and instead just pass whatever key & value we're sending in the API route

rathorevaibhav (approved):
Looking good! 👍

PR #51, opened 2022-12-05
IP whitelisting

Bahugunajii:
## Description
- Whitelisted all the required IP's

## Checklist 
- [x] Self review


lib/dbservice_web/endpoint.ex
```diff
@@ -44,5 +44,12 @@ defmodule DbserviceWeb.Endpoint do
   plug Plug.MethodOverride
   plug Plug.Head
   plug Plug.Session, @session_options
+
+  plug APIacFilterIPWhitelist,
+    whitelist: [
+      "127.0.0.1/32",
+      "35.190.11.178/32"
```
rathorevaibhav, line 51: is it possible to whitelist domains? instead of IPs? I think it would be better if we can control the domain names. They would be easy to read and update if needed

Bahugunajii: Whitelisting domain was a priority for me too but unfortunately after a lot of research, I was unable to find out any trace of domain whitelisting in Elixir...

config/dev.exs
```diff
@@ -58,3 +58,6 @@ config :phoenix, :stacktrace_depth, 20
 
 # Initialize plugs at runtime for faster development compilation
 config :phoenix, :plug_init_mode, :runtime
+
+# Setup configuration to fetch whitelisted ip's
+config :dbservice, :ip_whitelist, "127.0.0.1-127.0.0.1 35.190.11.178-35.190.11.178"
```
rathorevaibhav, line 63: what is the other IP? `35.190...`?

rathorevaibhav (approved):
one question. else looking good.

config/dev.exs
Bahugunajii: yes!

PR #59, opened 2023-02-20
Addition of additional parameter in query

Bahugunajii:
## Description
- Added a check for `response_size` parameter in the query which provides an optimized response if value of `response_size` is set to `compact`.

## Response comparison
- Output when response size = compact:
Screenshot 2023-02-21 121320

- Output without response size parameter:

Screenshot 2023-02-21 121409

## Checklist
- [x] Self review

### DIFF
--- a/lib/dbservice_web/controllers/student_controller.ex
+++ b/lib/dbservice_web/controllers/student_controller.ex
@@ -32,7 +32,13 @@ defmodule DbserviceWeb.StudentController do
   end
 
   def index(conn, params) do
-    param = Enum.map(params, fn {key, value} -> {String.to_existing_atom(key), value} end)
+    is_filtered = params["is_filtered"]
+    IO.inspect(is_filtered)
+    query_params = Map.delete(params, "is_filtered")
+    IO.inspect(query_params)
+    IO.inspect("hello")
+
+    param = Enum.map(query_params, fn {key, value} -> {String.to_existing_atom(key), value} end)
 
     student =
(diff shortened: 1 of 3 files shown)


lib/dbservice_web/controllers/student_controller.ex
```diff
@@ -32,7 +32,9 @@ defmodule DbserviceWeb.StudentController do
   end
 
   def index(conn, params) do
-    param = Enum.map(params, fn {key, value} -> {String.to_existing_atom(key), value} end)
+    query_params = Map.delete(params, "is_filtered")
```
rathorevaibhav, line 35: `is_filtered` looks like you are sending a filtered list of students. But instead you are not having any filters applied but rather have reduced columns. I would suggest using a different name for this param if you want to get reduced set of columns.

Maybe something like `response_size=compact` will work?

lib/dbservice_web/views/user_view.ex
```diff
@@ -28,4 +32,21 @@ defmodule DbserviceWeb.UserView do
       country: user.country
     }
   end
+
+  def render("optimized_user.json", %{user: user}) do
```
rathorevaibhav, line 36: this is not an optimized user. This is a user with reduced data fields. I would suggest renaming it to maybe `user_with_compact_fields.json` or something.

rathorevaibhav (changes requested):

rathorevaibhav (approved):

PR #77, opened 2023-04-06
Fix/unit tests

belwalshubham:
Implementation of Unit testing as per the new schema changes

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -60,3 +60,5 @@ config :dbservice, Dbservice.Repo,
 # Import environment specific config. This must remain at the bottom
 # of this file so it overrides the configuration defined above.
 import_config "#{config_env()}.exs"
+
+config :mixservice, env: Mix.env()
--- a/config/config.exs
+++ b/config/config.exs
@@ -61,4 +61,4 @@ config :dbservice, Dbservice.Repo,
 # of this file so it overrides the configuration defined above.
 import_config "#{config_env()}.exs"
 
-config :mixservice, env: Mix.env()
+config :dbservice, env: Mix.env()
--- a/lib/dbservice_web/controllers/user_controller.ex
+++ b/lib/dbservice_web/controllers/user_controller.ex
@@ -54,8 +54,8 @@ defmodule DbserviceWeb.UserController do
     response(201, "Created", Schema.ref(:User))
   end
 
-  def create(conn, params) do
-    with {:ok, %User{} = user} <- Users.create_user(params) do
+  def create(conn, %{"user" => user_params}) do
+    with {:ok, %User{} = user} <- Users.create_user(user_params) do
       conn
       |> put_status(:created)
       |> put_resp_header("location", Routes.user_path(conn, :show, user))
--- a/lib/dbservice_web/controllers/user_controller.ex
+++ b/lib/dbservice_web/controllers/user_controller.ex
@@ -24,7 +24,7 @@ defmodule DbserviceWeb.UserController do
   end
 
   swagger_path :index do
-    get("/api/user")
+    get("/api/user?full_name=Rahul Sharma")
     response(200, "OK", Schema.ref(:Users))
   end
 
@@ -67,7 +67,7 @@ defmodule DbserviceWeb.UserController do
     get("/api/user/{userId}")
 
     parameters do
-      userId(:path, :integer, "The id of the user", required: true)
+      userId(:path, :integer, "The id of the user record", required: true)
     end
 
     response(200, "OK", Schema.ref(:User))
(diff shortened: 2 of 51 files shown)


config/config.exs
```diff
@@ -60,3 +60,5 @@ config :dbservice, Dbservice.Repo,
 # Import environment specific config. This must remain at the bottom
 # of this file so it overrides the configuration defined above.
 import_config "#{config_env()}.exs"
+
+config :dbservice, env: Mix.env()
```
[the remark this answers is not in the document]
belwalshubham: After pushing the code for the domain_whitelist_plug.ex file, my checks on GitHub failed with the error message Function Mix.env/0 does not exist. So here is solution to avoid this error : https://stackoverflow.com/questions/35010950/get-current-environment-name/44747870#44747870

lib/dbservice_web/controllers/user_controller.ex
```diff
@@ -54,8 +54,8 @@ defmodule DbserviceWeb.UserController do
     response(201, "Created", Schema.ref(:User))
   end
 
-  def create(conn, params) do
-    with {:ok, %User{} = user} <- Users.create_user(params) do
+  def create(conn, %{"user" => user_params}) do
```
Bahugunajii, line 57: Why are we sending it as a struct?

We never send the parameters as a struct while making any post request.

Revert this!

lib/dbservice_web/views/session_occurence_view.ex
```diff
@@ -21,17 +19,4 @@ defmodule DbserviceWeb.SessionOccurenceView do
       session_id: session_occurence.session_id
     }
   end
```
Bahugunajii, line 21: Why you removed this piece of code? We need to show session_occurence_with_users no?

test/dbservice/users_test.exs
```diff
@@ -11,7 +11,7 @@ defmodule Dbservice.UsersTest do
     @invalid_attrs %{
       full_name: nil,
       email: nil,
-      phone: nil,
+      phone: "nope",
```
Bahugunajii, line 14: Rename this to something appropriate

test/dbservice/users_test.exs
```diff
@@ -11,7 +11,7 @@ defmodule Dbservice.UsersTest do
...
       gender: nil,
```
Bahugunajii, line 15: Also, I think let's remove phone from `validate_required` as there may be some cases where `phone` is null

test/dbservice_web/controllers/batch_controller_test.exs
```diff
@@ -0,0 +1,101 @@
...
+    end
+
+    test "renders errors when data is invalid", %{conn: conn} do
+      conn = post(conn, Routes.batch_path(conn, :create), batch: @invalid_attrs)
+      assert json_response(conn, 422)["errors"] != %{}
+    end
+  end
+
+  describe "update batch" do
+    setup [:create_batch]
```
Bahugunajii, line 61: why are we using `setup`. 
Will it work if we remove this?

test/dbservice_web/controllers/batch_controller_test.exs
```diff
@@ -0,0 +1,101 @@
...
+    end
+
+    test "renders errors when data is invalid", %{conn: conn, batch: batch} do
+      conn = put(conn, Routes.batch_path(conn, :update, batch), @invalid_attrs)
+      assert json_response(conn, 422)["errors"] != %{}
+    end
+  end
+
+  describe "delete batch" do
+    setup [:create_batch]
```
Bahugunajii, line 85: same here.
Also, why are we having `create_batch` under the `delete batch`

Bahugunajii (changes requested):

lib/dbservice_web/controllers/user_controller.ex
belwalshubham: removed

test/dbservice/users_test.exs
belwalshubham: That was already removed

test/dbservice_web/controllers/batch_controller_test.exs
belwalshubham: The setup block is being used to create a test batch record in the database before each test case that requires it. Specifically, the create_batch function is being called to create a new batch fixture and store it in the batch variable, which is then passed as an argument to the test case.

If the setup block is removed, the batch variable will not be set, and any test cases that depend on it will fail. Therefore, it is necessary to use the setup block to ensure that the required test data is available for each test case that needs it.

test/dbservice_web/controllers/batch_controller_test.exs
belwalshubham: In the provided code, the create_batch function is being used to create a batch record in the database before the delete batch test case is run. This is necessary because the delete batch test case expects a batch variable to be present in the context, which is used to specify the batch to be deleted.

Bahugunajii (approved):

PR #81, opened 2023-04-26
Add pagination to all index function

belwalshubham:
This pull request adds pagination to the index function in all controller. The offset and limit parameters are now used to control which subset of records is returned by the Ecto query, allowing clients to easily paginate through large result sets.

In addition, the function has been refactored to simplify the parameter handling and improve readability.

### DIFF
--- a/lib/dbservice_web/controllers/batch_controller.ex
+++ b/lib/dbservice_web/controllers/batch_controller.ex
@@ -23,18 +23,22 @@ defmodule DbserviceWeb.BatchController do
   end
 
   def index(conn, params) do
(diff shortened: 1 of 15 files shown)


lib/dbservice_web/controllers/batch_controller.ex
```diff
@@ -23,18 +23,22 @@ defmodule DbserviceWeb.BatchController do
...
-
-        _, query ->
-          query
+    query =
+      from m in Batch,
+        order_by: [asc: m.id],
+        offset: ^params["offset"],
+        limit: ^params["limit"]
+
+    query =
```
Bahugunajii, line 32: why are we using two variables with same name?

PR #83, opened 2023-05-01
Restore production database to local database

belwalshubham:
Script that fetches a data dump from a production PostgreSQL database and restores it to a local PostgreSQL database. It also removes the dump file after the restoration is complete.
 To run the script you need to run the following command in the terminal:
 `
 ./fetch-data.sh
 `

### DIFF
--- a/fetch-data.sh
+++ b/fetch-data.sh
@@ -15,7 +15,6 @@ local_db_name="dbservice_dev"
 local_db_user="postgres"
 local_db_password="postgres"
 
-
 # Dump file name
 dump_file="dump.sql"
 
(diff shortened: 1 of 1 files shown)


fetch-data.sh
```diff
@@ -0,0 +1,50 @@
...
+production_db_name="xxx"
+production_db_user="postgres"
+production_db_password="xxx"
+
+# Local database credentials
+local_db_host="localhost"
+local_db_name="dbservice_dev"
+local_db_user="postgres"
+local_db_password="postgres"
+
```
Bahugunajii, line 17: Let's add additional command to delete the existing tables of the local database first before creating dump in order to avoid remove irrelevant entries(if any) exisiting in the local db.

Bahugunajii (approved):

PR #104, opened 2023-06-15
Relationship between Session and Form Schema tables

sudheshna-donthineni:


Bahugunajii (changes requested):
Since we added `form_schema_id` in session, so we can also add it in `session_view` and also in swagger doc. Rest looks good!

Bahugunajii (approved):

PR #115, opened 2023-06-26
Configure number of fields in form

sudheshna-donthineni:


Bahugunajii:
seems like you only added `number_of_fields_in_pop_form` to session's schema. Where is the migration for that?

PR #139, opened 2023-10-09
Update session controller to enable pagination

suryabulusu:
This PR updates session controller GET request to enable the following:
- Get a list of sessions with `id` in descending order (using offset and limit)
- Get sessions with `session_id_is_null=true` 

So, a command like : `curl -X GET "http://localhost:4000/api/session?session_id_is_null=true&offset=0&limit=10" -H  "accept: application/json"` gives the latest ten sessions with id as null. 

This will be used by `quiz-creator` frontend to display sessions whose ID has not yet been generated. 

Additionally, this PR also updates local installation instructions for this repo.

### DIFF
--- a/lib/dbservice_web/controllers/session_controller.ex
+++ b/lib/dbservice_web/controllers/session_controller.ex
@@ -52,22 +52,21 @@ defmodule DbserviceWeb.SessionController do
 
     query =
       Enum.reduce(params, query, fn {key, value}, acc ->
-        case key do
-          "offset" ->
+        case String.to_existing_atom(key) do
+          :offset ->
             acc
 
-          "limit" ->
+          :limit ->
             acc
 
-          "sort_order" ->
+          :sort_order ->
             acc
 
-          "session_id_is_null" ->
+          :session_id_is_null ->
             apply_session_id_null_filter(value, acc)
 
-          _ ->
-            field_name = String.to_existing_atom(key)
-            from u in acc, where: field(u, ^field_name) == ^value
+          atom ->
+            from(u in acc, where: field(u, ^atom) == ^value)
         end
       end)
 
--- a/lib/dbservice_web/controllers/session_controller.ex
+++ b/lib/dbservice_web/controllers/session_controller.ex
@@ -42,12 +42,7 @@ defmodule DbserviceWeb.SessionController do
   end
 
   def index(conn, params) do
-    sort_order =
-      case params["sort_order"] do
-        "asc" -> :asc
-        # default
-        _ -> :desc
-      end
+    sort_order = extract_sort_order(params)
 
     query =
       from m in Session,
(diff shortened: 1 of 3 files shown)


lib/dbservice_web/controllers/session_controller.ex
```diff
@@ -44,16 +44,25 @@ defmodule DbserviceWeb.SessionController do
   def index(conn, params) do
     query =
       from m in Session,
-        order_by: [asc: m.id],
+        order_by: [desc: m.id],
```
Bahugunajii, line 47: Let's make this ordering dynamic. We can add `sort_order` in query param such that if `sort_order` is `asc`, the response will be sorted in ascending order. If `sort_order` is `desc`, result will be sorted in descending order. If `sort_order` in not provided in query, then by default it will be sorted in descending.

Bahugunajii (changes requested):

lib/dbservice_web/controllers/session_controller.ex
```diff
@@ -42,18 +42,40 @@ defmodule DbserviceWeb.SessionController do
...
+          "limit" ->
+            acc
+
+          "sort_order" ->
+            acc
+
+          "session_id_is_null" ->
+            case value do
+              "true" -> from u in acc, where: is_nil(u.session_id)
+              "false" -> from u in acc, where: not is_nil(u.session_id)
```
Bahugunajii, line 73: What's the use case of this condition?

suryabulusu: i need two tables on frontend... one where session id is null .. and one where it is not null

lib/dbservice_web/controllers/session_controller.ex
```diff
@@ -42,18 +42,40 @@ defmodule DbserviceWeb.SessionController do
...
+              _ -> acc
+            end
+
+          _ ->
+            from u in acc, where: fragment("? = ?", field(u, ^key), ^value)
```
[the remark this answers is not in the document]
Bahugunajii: Then let's use `String.to_existing_atom`. This looks weird

lib/dbservice_web/controllers/session_controller.ex
```diff
@@ -42,25 +42,54 @@ defmodule DbserviceWeb.SessionController do
...
+
+          "sort_order" ->
+            acc
+
+          "session_id_is_null" ->
+            apply_session_id_null_filter(value, acc)
+
+          _ ->
+            field_name = String.to_existing_atom(key)
+            from u in acc, where: field(u, ^field_name) == ^value
```
Bahugunajii, line 70:
```suggestion
                case String.to_existing_atom(key) do
          :offset ->
            acc

          :limit ->
            acc

          :sort_order ->
            acc

          :session_id_is_null ->
            apply_session_id_null_filter(value, acc)

          atom ->
            from(u in acc, where: field(u, ^atom) == ^value)
        end
      end)
```

Can you try this and check if it is working for all cases?

Bahugunajii (changes requested):

Bahugunajii (approved):

PR #142, opened 2023-11-30
Changing from uuid to teacher_id

sudheshna-donthineni:


suryabulusu (approved):
looks good.. where is migration file

suryabulusu (approved):

lib/dbservice/users/teacher.ex
```diff
@@ -11,7 +11,7 @@ defmodule Dbservice.Users.Teacher do
     field :designation, :string
     field :grade, :string
     field :subject, :string
-    field :uuid, :string
+    field :teacher_id, :string
```
pritamps, line 14: Don't remove `uuid` -- We can't guarantee uniqueness of `teacher_id` because maybe we'll set it manually or something. Have both `uuid` (system generated) and `teacher_id`

pritamps (changes requested):

lib/dbservice/users/teacher.ex
sudheshna-donthineni: `uuid` is not system generated, system generated `id` is different. This is the ID for a teacher that we give, just like `student_id`. I'm just renaming it for better understanding.

PR #144, opened 2024-01-03
Class Library

Bahugunajii:
### Description
- Added `batch_id` and `parent_id` to batch
- Added `teacher_id` to resource

### Checklist
- [x] Self-testing


lib/dbservice_web/plugs/authentication_middleware_plug.ex
```diff
@@ -35,7 +35,7 @@ defmodule DbserviceWeb.AuthenticationMiddleware do
     end
   end
 
-  defp is_swagger_request?(referer, request_path) do
+  defp swagger_request?(referer, request_path) do
```
sudheshna-donthineni, line 38: Can you double check this function name? In my PR, this was the issue, i renamed it to `contain_swagger_request`.

lib/dbservice/batches/batch.ex
```diff
@@ -10,6 +10,7 @@ defmodule Dbservice.Batches.Batch do
   schema "batch" do
     field :name, :string
     field :contact_hours_per_week, :integer
+    field :batch_id, :string
```
sudheshna-donthineni, line 13: Remember how we talked about having a parent batch as well? Maybe you can add that change to this PR only, since you are anyway altering the batch table.

lib/dbservice_web/plugs/authentication_middleware_plug.ex
Bahugunajii: okay, let it be `contain_swagger_request`

lib/dbservice/batches/batch.ex
Bahugunajii: The migration created for this has already ran in staging. Let's make a separate PR for this

sudheshna-donthineni (approved):

PR #145, opened 2024-01-11
Removed has_internet_access column from session table

sudheshna-donthineni:


Bahugunajii (changes requested):
Let's also remove it from schema file, view and swagger.

Bahugunajii (approved):

PR #147, opened 2024-01-27
Creation of new resource or updation of resources is now dependant on the db-service

sudheshna-donthineni:
The `db-service` checks:
-  if a new session needs to be created or an existing session can be updated based on the session ID.
- if a new student needs to be created or an existing student can be updated based on the student ID.
-  if a new group-session needs to be created or an existing group-session can be updated based on the session ID.
-  if a new batch needs to be created or an existing batch can be updated based on the batch ID.
- if a new teacher needs to be created or an existing teacher can be updated based on the teacher ID.
- if a new exam needs to be created or an existing exam can be updated based on the exam name.
- if a new group needs to be created or an existing group can be updated based on the group name.

### DIFF
--- a/lib/dbservice/batches.ex
+++ b/lib/dbservice/batches.ex
@@ -31,6 +31,19 @@ defmodule Dbservice.Batches do
   def get_batch!(id), do: Repo.get!(Batch, id)
 
   @doc """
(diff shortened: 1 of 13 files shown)


suryabulusu (approved):
looks good

PR #148, opened 2024-02-01
Allow session deletion

suryabulusu:
Deleting a session also involves
- Deleting corresponding entry in GroupSession table
- Deleting corresponding entries in SessionOccurrence table

These are handled outside db-service as of now


Bahugunajii (changes requested):
There might be a case where other table has referenced the `session_id` which we want to delete. In that case it will always throw a `foreign_key_constraint`. So let's add a functionality  to solve this issue.

Bahugunajii (approved):

Bahugunajii (approved):

PR #150, opened 2024-02-14
Session schedule schema

Bahugunajii:
### Description
- Session schedule schema

### Checklist
- [x] Self-testing

### DIFF
--- a/lib/dbservice/batches/batch.ex
+++ b/lib/dbservice/batches/batch.ex
@@ -6,6 +6,7 @@ defmodule Dbservice.Batches.Batch do
 
   alias Dbservice.Programs.Program
   alias Dbservice.Groups.GroupType
+  alias Dbservice.Sessions.SessionSchedule
 
   schema "batch" do
     field :name, :string
@@ -15,6 +16,7 @@ defmodule Dbservice.Batches.Batch do
 
     has_many :group_type, GroupType, foreign_key: :child_id, where: [type: "batch"]
     many_to_many :program, Program, join_through: "batch_program", on_replace: :delete
+    has_one :session_schedule, SessionSchedule
 
     timestamps()
   end
--- a/lib/dbservice/session_schedules.ex
+++ b/lib/dbservice/session_schedules.ex
@@ -1,0 +1,81 @@
+defmodule Dbservice.SessionSchedules do
+  @moduledoc """
+  The SessionSchedules context.
+  """
+
+  import Ecto.Query, warn: false
+  alias Dbservice.Repo
+
+  alias Dbservice.Sessions.SessionSchedule
+
+  @doc """
+  Returns the list of session_schedule.
+  ## Examples
+      iex> list_session_schedule()
+      [%SessionSchedule{}, ...]
+  """
+  def list_session_schedule do
+    Repo.all(SessionSchedule)
+  end
+
+  @doc """
+  Gets a single session_schedule.
+  Raises `Ecto.NoResultsError` if the session_schedule does not exist.
+  ## Examples
+      iex> get_session_schedule!(123)
+      %SessionSchedule{}
+      iex> get_session_schedule!(456)
+      ** (Ecto.NoResultsError)
+  """
+  def get_session_schedule!(id), do: Repo.get!(SessionSchedule, id)
+
+  @doc """
+  Creates a session_schedule.
+  ## Examples
+      iex> create_session_schedule(%{field: value})
+      {:ok, %SessionSchedule{}}
+      iex> create_session_schedule(%{field: bad_value})
+      {:error, %Ecto.Changeset{}}
+  """
+  def create_session_schedule(attrs \\ %{}) do
+    %SessionSchedule{}
+    |> SessionSchedule.changeset(attrs)
+    |> Repo.insert()
+  end
+
+  @doc """
+  Updates a session_schedule.
+  ## Examples
+      iex> update_session_schedule(session_schedule, %{field: new_value})
+      {:ok, %SessionSchedule{}}
+      iex> update_session_schedule(session_schedule, %{field: bad_value})
+      {:error, %Ecto.Changeset{}}
+  """
+  def update_session_schedule(%SessionSchedule{} = session_schedule, attrs) do
+    session_schedule
+    |> SessionSchedule.changeset(attrs)
+    |> Repo.update()
+  end
+
+  @doc """
+  Deletes a session_schedule.
+  ## Examples
+      iex> delete_session_schedule(session_schedule)
+      {:ok, %SessionSchedule{}}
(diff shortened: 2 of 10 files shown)


lib/dbservice/batches/batch.ex
```diff
@@ -15,6 +16,7 @@ defmodule Dbservice.Batches.Batch do
 
     has_many :group_type, GroupType, foreign_key: :child_id, where: [type: "batch"]
     many_to_many :program, Program, join_through: "batch_program", on_replace: :delete
+    has_one :session_schedule, SessionSchedule
```
pritamps, line 19: This looks so nice :) Batch has one session schedule. Makes sense in English also :)

lib/dbservice/session_schedules.ex
```diff
@@ -0,0 +1,81 @@
...
+  @doc """
+  Gets a single session_schedule.
+  Raises `Ecto.NoResultsError` if the session_schedule does not exist.
+  ## Examples
+      iex> get_session_schedule!(123)
+      %SessionSchedule{}
+      iex> get_session_schedule!(456)
+      ** (Ecto.NoResultsError)
+  """
+  def get_session_schedule!(id), do: Repo.get!(SessionSchedule, id)
```
pritamps, line 30: `id` here refers to the internal db id, right?

Can we add a method to get session schedule based on batch id?

lib/dbservice/sessions/session.ex
```diff
@@ -36,6 +37,7 @@ defmodule Dbservice.Sessions.Session do
     many_to_many(:users, User, join_through: "user_session", on_replace: :delete)
     many_to_many(:group_type, GroupType, join_through: "group_session", on_replace: :delete)
     belongs_to(:form_schema, FormSchema)
+    has_many(:session_schedule, SessionSchedule)
```
pritamps, line 40: What is the use case for one session having multiple session schedules?

lib/dbservice_web/controllers/session_schedule_controller.ex
```diff
@@ -0,0 +1,131 @@
...
+      params(:query, :string, "The id of a session",
+        required: false,
+        name: "id"
+      )
+    end
+
+    response(200, "OK", Schema.ref(:SessionSchedules))
+  end
+
+  def index(conn, params) do
```
pritamps, line 35: If we have days of week and start and end time, can we store them so that it doesn't require any reformatting on client end? Maybe not, actually -- but something to think about

lib/dbservice_web/controllers/session_schedule_controller.ex
```diff
@@ -0,0 +1,131 @@
...
+    today_day_of_week = Calendar.Date.day_of_week_name(Date.utc_today())
+
+    query =
```
pritamps, line 38: Sorry, maybe I'm not understanding elixir code ;)

but why not filter by batch ID also here? 

The main (and only, for now) use case is to get the current session schedule for a batch, right?

lib/dbservice_web/controllers/session_schedule_controller.ex
```diff
@@ -0,0 +1,131 @@
...
+      |> put_resp_header("location", Routes.session_schedule_path(conn, :show, session_schedule))
+      |> render("show.json", session_schedule: session_schedule)
+    end
+  end
+
+  swagger_path :show do
+    get("/api/session-schedule/{sessionScheduleId}")
+
+    parameters do
+      sessionScheduleId(:path, :integer, "The id of the session schedule record", required: true)
```
pritamps, line 83: We need one to be able to retrieve session schedule for a batch (most common use case)

lib/dbservice_web/controllers/session_schedule_controller.ex
```diff
@@ -0,0 +1,131 @@
...
+  end
+
+  def delete(conn, %{"id" => id}) do
+    session_schedule = SessionSchedules.get_session_schedule!(id)
+
+    with {:ok, %SessionSchedule{}} <- SessionSchedules.delete_session_schedule(session_schedule) do
+      send_resp(conn, :no_content, "")
+    end
+  end
+end
```
pritamps, line 131: just clarifying my understanding here: Currently creation of session schedule here is through a API that we can't really use now. This API will be called by something external to DB service when sessions or their schedules are created/modified (for example, on Google Calendar?)

pritamps (changes requested):

lib/dbservice/session_schedules.ex
Bahugunajii: It's already implemented. We can just use `batch_id` as a query parameter something like:
`/api/session-schedule?batch_id=1`

lib/dbservice_web/controllers/session_schedule_controller.ex
Bahugunajii: `batch_id` is being sent as a query parameter from client side. In db, let's say if we have 12 sessions (for different days and times) belonging to a same `batch_id`, then this will return only the session happening today.

lib/dbservice_web/controllers/session_schedule_controller.ex
Bahugunajii: Already implemented. So there are 5 types of API endpoints:
1) Index API (GET /api/session-schedule):
- Purpose: Retrieve a list of session schedules
- Query parameters can be anything we want to filter (like `batch_id`, `session_id`, `start_time` etc. etc.)

2) Create API (POST /api/session-schedule):
- Purpose: Create a new session schedule.

3) Show API (GET /api/session-schedule/{sessionScheduleId}):
- Purpose: Retrieve details of a specific session schedule based on id (PK) of session schedule

4) Update API (PATCH /api/session-schedule/{sessionScheduleId}):
- Purpose: Update an existing session schedule.

5) Delete API (DELETE /api/session-schedule/{sessionScheduleId}):
- Purpose: Delete a session schedule.

This is the 3rd one listed above. If we want to filter session schedule based on batch, we will use the first one (Index API) by sending `batch_id` as query param.

lib/dbservice_web/controllers/session_schedule_controller.ex
Bahugunajii: Correct!

lib/dbservice/sessions/session.ex
Bahugunajii: If one session is happening for different batches and the schedule for each batch is different, then I think one to many will work

lib/dbservice_web/controllers/session_schedule_controller.ex
Bahugunajii: I'm sorry I didn't get the point. Can you please explain?

pritamps (approved):

PR #151, opened 2024-02-16
Remove :register APIs for student and teacher

sudheshna-donthineni:
Now, db-service handles the creation of a new student/teacher. It first checks if the student/teacher already exists, if yes, then updates the student/teacher (along with user), otherwise creates a student/teacher (along with user).
Since student and teacher are coupled with user, any updating of student and teacher should also update the user.


lib/dbservice_web/controllers/student_controller.ex
```diff
@@ -135,26 +135,6 @@ defmodule DbserviceWeb.StudentController do
...
-    post("/api/student/register")
-
-    parameters do
-      body(:body, Schema.ref(:StudentRegistration), "Student to create along with user",
-        required: true
-      )
-    end
-
-    response(201, "Created", Schema.ref(:StudentWithUser))
-  end
```
Bahugunajii, line 148: Why swagger path for student registration is removed?

pritamps (approved):

lib/dbservice_web/controllers/student_controller.ex
```diff
@@ -76,7 +76,7 @@ defmodule DbserviceWeb.StudentController do
   def create(conn, params) do
     case Users.get_student_by_student_id(params["student_id"]) do
       nil ->
-        create_new_student(conn, params)
+        create_student_with_user(conn, params)
 
       existing_student ->
         update_existing_student(conn, existing_student, params)
```
Bahugunajii, line 82: This should be `update_existing_student_with_user`

lib/dbservice_web/controllers/student_controller.ex
```diff
@@ -166,19 +146,18 @@ defmodule DbserviceWeb.StudentController do
     end
   end
 
-  defp create_new_student(conn, params) do
-    with {:ok, %Student{} = student} <- Users.create_student(params) do
+  defp update_existing_student(conn, existing_student, params) do
+    with {:ok, %Student{} = student} <- Users.update_student(existing_student, params) do
```
Bahugunajii, line 150: Use `Users.update_student_with_user` instead of `update_student`.

lib/dbservice_web/controllers/teacher_controller.ex
```diff
@@ -132,26 +132,6 @@ defmodule DbserviceWeb.TeacherController do
...
-    post("/api/teacher/register")
-
-    parameters do
-      body(:body, Schema.ref(:TeacherRegistration), "Teacher to create along with user",
-        required: true
-      )
-    end
-
-    response(201, "Created", Schema.ref(:TeacherWithUser))
-  end
```
Bahugunajii, line 145: Also why swagger path for teacher registration is removed?

lib/dbservice_web/controllers/teacher_controller.ex
```diff
@@ -73,7 +73,7 @@ defmodule DbserviceWeb.TeacherController do
   def create(conn, params) do
     case Users.get_teacher_by_teacher_id(params["teacher_id"]) do
       nil ->
-        create_new_teacher(conn, params)
+        create_teacher_with_user(conn, params)
 
       existing_teacher ->
         update_existing_teacher(conn, existing_teacher, params)
```
Bahugunajii, line 79: Similarly, update existing teacher with user.

Bahugunajii (changes requested):

lib/dbservice_web/controllers/student_controller.ex
sudheshna-donthineni: We already have a swagger path for create, no.

lib/dbservice_web/controllers/student_controller.ex
sudheshna-donthineni: It's only updating student. This is still the same as before.

Bahugunajii: That's what I'm saying. It shouldn't only update student. Rather it should now update student along with its user!

sudheshna-donthineni: Okay, read the next reviews and got it.

lib/dbservice_web/controllers/student_controller.ex
Bahugunajii: Ideally this should now be:
`
swagger_path :register do
    post("/api/student")

    parameters do
      body(:body, Schema.ref(:StudentRegistration), "Student to create along with user",
        required: true
      )
    end

    response(201, "Created", Schema.ref(:StudentWithUser))
  end
`

The Schema references should change.

lib/dbservice_web/controllers/student_controller.ex
Bahugunajii: So, let's make the changes then?

lib/dbservice_web/controllers/teacher_controller.ex
Bahugunajii:
```
  swagger_path :create do
    post("/api/teacher")

    parameters do
      body(:body, Schema.ref(:TeacherRegistration), "Teacher to create along with user",
        required: true
      )
    end

    response(201, "Created", Schema.ref(:TeacherWithUser))
  end
  ```

**Review by Bahugunajii — CHANGES_REQUESTED** [review]
Similarly, for update and delete functions, we will be using `update_student_with_user` and `delete_student_with_user`.

Let's make these slight changes!

**Comment by sudheshna-donthineni** [comment]
> Similarly, for update and delete functions, we will be using `update_student_with_user` and `delete_student_with_user`.
> 
> Let's make these slight changes!

`update_student_with_user` is already implemented. We technically don't want to have deletion functionality.

File: lib/dbservice_web/controllers/student_controller.ex [hunk 6]
```diff
@@ -67,19 +67,19 @@ defmodule DbserviceWeb.StudentController do
     post("/api/student")
 
     parameters do
-      body(:body, Schema.ref(:Student), "Student to create", required: true)
+      body(:body, Schema.ref(:Student), "Student to create along with user", required: true)
```
Bahugunajii, line 70:
```suggestion
      body(:body, Schema.ref(:StudentWithUser), "Student to create along with user", required: true)
```

lib/dbservice_web/controllers/teacher_controller.ex
```diff
@@ -64,19 +64,19 @@ defmodule DbserviceWeb.TeacherController do
     post("/api/teacher")
 
     parameters do
-      body(:body, Schema.ref(:Teacher), "Teacher to create", required: true)
+      body(:body, Schema.ref(:Teacher), "Teacher to create along with user", required: true)
```
Bahugunajii, line 67:
```suggestion
      body(:body, Schema.ref(:TeacherWithUser), "Teacher to create along with user", required: true)
```

Bahugunajii (changes requested):

Bahugunajii:
> > Similarly, for update and delete functions, we will be using `update_student_with_user` and `delete_student_with_user`.
> > Let's make these slight changes!
> 
> `update_student_with_user` is already implemented. We technically don't want to have deletion functionality.

What I'm trying to say is, we have a PATCH /student API and the fucntion definition looks like:
```
  def update(conn, params) do
    student = Users.get_student!(params["id"])

    with {:ok, %Student{} = student} <- Users.update_student(student, params) do
      render(conn, "show.json", student: student)
    end
  end
  ```
  
  So instead of updating only the student here, we can update it to update student along with the user!

Bahugunajii (approved):

PR #167, opened 2024-04-08
Feat: API to retrieve sessions belonging to a user

Bahugunajii:
### Description
- Now we have a support of an API which will allow us to fetch all the live classes and quiz sessions belonging to the logged in user in gurukul

### Checklist
- [x] Self-testing

### DIFF
--- a/lib/dbservice/groups.ex
+++ b/lib/dbservice/groups.ex
@@ -32,6 +32,19 @@ defmodule Dbservice.Groups do
   def get_group!(id), do: Repo.get!(Group, id)
 
   @doc """
+  Gets a single group.
+  Raises `Ecto.NoResultsError` if the Group does not exist.
+  ## Examples
+      iex> get_group_by_group_id(123)
+      %Group{}
+      iex> get_group_by_group_id(456)
+      ** (Ecto.NoResultsError)
+  """
+  def get_group_by_group_id(group_id) do
+    Repo.get_by(Group, id: group_id, type: "batch")
+  end
+
+  @doc """
   Creates a group.
   ## Examples
       iex> create_group(%{field: value})
--- a/lib/dbservice/groups.ex
+++ b/lib/dbservice/groups.ex
@@ -32,7 +32,7 @@ defmodule Dbservice.Groups do
   def get_group!(id), do: Repo.get!(Group, id)
 
   @doc """
-  Gets a single group.
+  Gets group by group_id.
   Raises `Ecto.NoResultsError` if the Group does not exist.
   ## Examples
       iex> get_group_by_group_id(123)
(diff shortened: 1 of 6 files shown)


lib/dbservice/groups.ex
```diff
@@ -31,6 +31,32 @@ defmodule Dbservice.Groups do
...
+  Gets group by group_id.
+  Raises `Ecto.NoResultsError` if the Group does not exist.
+  ## Examples
+      iex> get_group_by_group_id(123)
+      %Group{}
+      iex> get_group_by_group_id(456)
+      ** (Ecto.NoResultsError)
+  """
+  def get_group_by_group_id(group_id) do
+    Repo.get_by(Group, id: group_id, type: "batch")
```
suryabulusu, line 44: why hardcode `batch` here? maybe this can be in input too?

lib/dbservice/groups.ex
```diff
@@ -31,6 +31,32 @@ defmodule Dbservice.Groups do
...
+  Gets group by child_id.
+  Raises `Ecto.NoResultsError` if the Group does not exist.
+  ## Examples
+      iex> get_group_by_child_id(123)
+      %Group{}
+      iex> get_group_by_child_id(456)
+      ** (Ecto.NoResultsError)
+  """
+  def get_group_by_child_id(child_id) do
+    Repo.get_by(Group, child_id: child_id, type: "batch")
```
suryabulusu, line 57: same comment as above

lib/dbservice_web/controllers/user_controller.ex
```diff
@@ -170,4 +176,45 @@ defmodule DbserviceWeb.UserController do
...
+
+    case Groups.get_group_by_group_id(group_id) do
+      nil -> []
+      group -> get_group(group, quiz_flag)
+    end
+  end
+
+  defp get_group(group, quiz_flag) do
+    child_id = group.child_id
+    batch = Batches.get_batch!(child_id)
```
suryabulusu, line 203: here we assume that type is always batch.. maybe its fine for now but later on we will have to expand to other types

suryabulusu (commented):
left comments

lib/dbservice/groups.ex
Bahugunajii: Currently we always have sessions for a batch. Later if there is a need where session are associated with a program, product, etc. then we can have a more generic function where type can be just passed as an argument of this function.

suryabulusu: got it

suryabulusu: i'm making an issue so that we can remember this for future

suryabulusu (approved):
looks good. made an issue on repo for future reference

PR #172, opened 2024-05-15
Removes current batch/grade/group

suryabulusu:
Redundant info in user profile


mix.lock
Bahugunajii, line 1: Let's not add `mix.lock` file as it may throw version compatibility errors while deployment

Bahugunajii (changes requested):

pritamps (approved):
Looks good!

pritamps (approved):

Bahugunajii (approved):

PR #175, opened 2024-06-12
Add upsert functionality:

Abhinav-3009:
- Modified program_controller to check if program exists, then update or create

- Modified product_controller to check if product exists, then update or create

### DIFF
--- a/lib/dbservice/programs.ex
+++ b/lib/dbservice/programs.ex
@@ -4,6 +4,7 @@ defmodule Dbservice.Programs do
   """
 
   import Ecto.Query, warn: false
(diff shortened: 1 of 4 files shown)


lib/dbservice/programs.ex
```diff
@@ -4,6 +4,7 @@ defmodule Dbservice.Programs do
   """
 
   import Ecto.Query, warn: false
+  alias DbserviceWeb.SwaggerSchema.Program
```
Bahugunajii, line 7: Same question as above

Bahugunajii (changes requested):

PR #177, opened 2024-06-18
Feat: Funtionality mark student as enrolled

Bahugunajii:
### Description
- Added an enpoint which:
    - Creates an entry in group-user for the current batch enrollment
    - Creates two entries in the enrollment record: one for the current batch enrollment and one for the current status
    - Set's student status to "enrolled"

### Checklist
- [x] Self-testing

### DIFF
--- a/lib/dbservice_web/controllers/student_controller.ex
+++ b/lib/dbservice_web/controllers/student_controller.ex
@@ -211,7 +211,7 @@ defmodule DbserviceWeb.StudentController do
         user_id: user_id,
         is_current: true,
         start_date: current_time,
-        status_group_id: status_group_id,
+        group_id: status_group_id,
         group_type: group_type,
         academic_year: academic_year,
         grade_id: grade_id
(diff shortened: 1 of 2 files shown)


lib/dbservice_web/controllers/student_controller.ex
```diff
@@ -204,12 +206,12 @@ defmodule DbserviceWeb.StudentController do
...
       end)
 
-      # Create a new enrollment record with the fetched group_id
+      # Create a new enrollment record with the fetched status_group_id
       new_enrollment_attrs = %{
         user_id: user_id,
         is_current: true,
         start_date: current_time,
-        group_id: group_id,
+        status_group_id: status_group_id,
```
sudheshna-donthineni, line 214: This won't work, I think. It should be group_id only, because ER has group_id as attribute.

lib/dbservice_web/controllers/student_controller.ex
```diff
@@ -228,4 +230,100 @@ defmodule DbserviceWeb.StudentController do
       end
     end
   end
+
+  def enrolled(conn, params) do
```
sudheshna-donthineni, line 234: I think you should write comments for this function.

lib/dbservice_web/controllers/student_controller.ex
Bahugunajii: Correct. The key won't match

Bahugunajii: Changing this

sudheshna-donthineni (approved):

PR #181, opened 2024-06-24
Feat: Filter quiz sessions based on batch ID

Bahugunajii:
### ADR
https://www.notion.so/avantifellows/ADR-Class-Batch-Metadata-for-Quiz-Sessions-cd84df30f7db46ab91bf023ea6aa9e68

### Description
- Filter quiz sessions based on the `batch_id` in metadata

### Checklist
- [x] Self-testing

### DIFF
--- a/lib/dbservice_web/controllers/user_controller.ex
+++ b/lib/dbservice_web/controllers/user_controller.ex
@@ -192,7 +192,7 @@ defmodule DbserviceWeb.UserController do
   defp get_group_user(group_user, quiz_flag) do
     group_id = group_user.group_id
 
-    case Groups.get_group_by_group_id(group_id) do
+    case Groups.get_group_by_group_id_and_type(group_id, "batch") do
       nil -> []
       group -> get_group(group, quiz_flag)
     end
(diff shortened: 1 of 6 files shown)


lib/dbservice_web/controllers/user_controller.ex
```diff
@@ -201,14 +201,42 @@ defmodule DbserviceWeb.UserController do
   defp get_group(group, quiz_flag) do
     child_id = group.child_id
     batch = Batches.get_batch!(child_id)
+    class_batch_id = batch.batch_id
     quiz_id = batch.parent_id
```
suryabulusu, line 205: quiz_batch _id

lib/dbservice_web/controllers/user_controller.ex
```diff
@@ -201,14 +201,42 @@ defmodule DbserviceWeb.UserController do
...
     quiz_group = Groups.get_group_by_child_id(quiz_id)
```
suryabulusu, line 206: quiz_batch_group

lib/dbservice_web/controllers/user_controller.ex
```diff
@@ -201,14 +201,42 @@ defmodule DbserviceWeb.UserController do
...
+
+      _ ->
+        if quiz_flag do
+          if !is_nil(session.meta_data["batch_id"]) do
+            session.meta_data["batch_id"] == class_batch_id and session.platform == "quiz"
+          else
+            false
+          end
+        else
+          session.platform == "meet"
```
suryabulusu, line 237: why this? you can just put true maybe? if quiz flag is false then eveyr other platform is allowed. not just meet

suryabulusu (approved):
left comments

lib/dbservice_web/controllers/user_controller.ex
Bahugunajii: But for live classes, shouldn't the platform be always meet?

Bahugunajii: In Gurukul, we show only Live classes and Quiz. My intent was to show "meet" sessions for live classes and "quiz" sessions for Tests.

lib/dbservice_web/controllers/user_controller.ex
Bahugunajii: But quiz_id is PK here

PR #185, opened 2024-06-28
Feat: Render session details with session occurrences

Bahugunajii:
### Description
- Render session details with session occurrences
- Fixed spelling

### Checklist
- [x] Self-testing

### DIFF
--- a/lib/dbservice/sessions/session.ex
+++ b/lib/dbservice/sessions/session.ex
@@ -7,6 +7,7 @@ defmodule Dbservice.Sessions.Session do
   alias Dbservice.Users.User
   alias Dbservice.Groups.Group
   alias Dbservice.Sessions.SessionSchedule
+  alias Dbservice.Sessions.SessionOccurrence
 
   schema "session" do
     field(:end_time, :utc_datetime)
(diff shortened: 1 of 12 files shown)


lib/dbservice/sessions/session.ex
```diff
@@ -37,6 +38,7 @@ defmodule Dbservice.Sessions.Session do
     many_to_many(:users, User, join_through: "user_session", on_replace: :delete)
     many_to_many(:group, Group, join_through: "group_session", on_replace: :delete)
     has_many(:session_schedule, SessionSchedule)
+    has_many(:session_occurrence, SessionOccurrence)
```
suryabulusu, line 41: why.. this will be too big. live classes have 365 session occurrences for example

suryabulusu (commented):
left comment.. unsure if we have to return all session occurrences along with session. 

in session manager, we retrieve 50 sessions at once on each page. The request would be too large if we get all their corresponding occurrences

lib/dbservice/sessions/session.ex
Bahugunajii: It's opposite. We load session details along with session occurrence. It will be used in gurukul

suryabulusu (approved):

PR #187, opened 2024-07-01
added create_student_id endpoint with its helper functions in student…

Abhinav-3009:
…_controller with its router

PR #190, opened 2024-07-08
used join instead of nested loop

Abhinav-3009:
### Problem Statement
In the PR: [added create_student_id endpoint with its helper functions in student… #187](https://github.com/avantifellows/db-service/pull/187)
We were using nested loops:
-> First we get existing students with `grade` and `category` from parameters.
-> Then we run a loop for each existing student to get existing user with `date_of_birth`, `gender`, `first_name` from parameters.
-> Then we loop to find enrollment record for each user if they are enrolled in the school specified in parameters.
### Result
This resulted in high response time(more than 30 seconds) of generating a student id or fetching an existing student id.

### Solution
We have now used a join between `student` table and `user` table to reduce the high response time.

### DIFF
--- a/lib/dbservice/users.ex
+++ b/lib/dbservice/users.ex
@@ -463,4 +463,23 @@ defmodule Dbservice.Users do
     |> where(^Util.build_conditions(params))
     |> Repo.all()
   end
+
+  @doc """
+  Gets a list of students along with their associated users based on the given parameters.
+  Returns an empty list if no matching students or users are found.
+  """
+  def get_students_with_users(grade_id, category, date_of_birth, gender, first_name) do
+    from(s in Student,
+      join: u in User,
+      on: u.id == s.user_id,
+      where:
+        s.grade_id == ^grade_id and
+          s.category == ^category and
+          u.date_of_birth == ^date_of_birth and
+          u.gender == ^gender and
+          u.first_name == ^first_name,
+      select: {s, u}
+    )
+    |> Repo.all()
+  end
 end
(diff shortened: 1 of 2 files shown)


lib/dbservice_web/controllers/student_controller.ex
```diff
@@ -438,47 +444,30 @@ defmodule DbserviceWeb.StudentController do
...
+      student_id
     else
-      [school] = Schools.get_school_by_params(%{name: params["school_name"]})
-
-      if check_existing_enrollment(existing_user, school.id) do
-        student_id
-      else
-        nil
-      end
+      nil
```
Bahugunajii, line 458: What will be the response if school is not there in the db?

lib/dbservice/users.ex
```diff
@@ -463,4 +463,23 @@ defmodule Dbservice.Users do
...
+      where:
+        s.grade_id == ^grade_id and
+          s.category == ^category and
+          u.date_of_birth == ^date_of_birth and
+          u.gender == ^gender and
+          u.first_name == ^first_name,
+      select: {s, u}
+    )
+    |> Repo.all()
+  end
```
Bahugunajii, line 484: Is there any possibility to use build_conditions func. here?

Abhinav-3009: I tried using build_conditions function here as well, but it didn't work.
build_conditions expects a single table to build where conditions but here we are using join in two tables so it gives an error.

Bahugunajii (approved):

PR #197, opened 2024-07-16
Bugfix: Fix user-session timestamp validation for IST

Bahugunajii:
### Description
- Updated `invalidate_future_date` function in Utils module to use DateTime for IST conversions and comparisons.

### Checklist
- [x] Local testing

### DIFF
--- a/lib/dbservice/utils/util.ex
+++ b/lib/dbservice/utils/util.ex
@@ -6,10 +6,20 @@ defmodule Dbservice.Utils.Util do
   import Ecto.Query
 
   def invalidate_future_date(changeset, date_field_atom) do
-    today = Date.utc_today()
+    utc_now = DateTime.utc_now()
+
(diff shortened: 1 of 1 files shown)


lib/dbservice/utils/util.ex
```diff
@@ -6,10 +6,13 @@ defmodule Dbservice.Utils.Util do
   import Ecto.Query
 
   def invalidate_future_date(changeset, date_field_atom) do
-    today = Date.utc_today()
+    utc_now = DateTime.utc_now()
+
+    ist_now = DateTime.add(utc_now, 5 * 60 * 60 + 30 * 60, :second)
```
pritamps, line 11: Maybe use the `Timex` library instead of doing it manually? 

https://stackoverflow.com/questions/62443581/what-is-the-correct-way-of-shifting-the-timezone-elixir

suryabulusu (approved):

PR #198, opened 2024-07-18
Get auth group from session id

Abhinav-3009:
### Summary

- This PR moves the logic for retrieving auth group data from the portal-backend to the db-service. This change aims to optimize our API structure by reducing multiple API calls to a single call.

### Changes

- Implemented new endpoint in db-service: `GET /api/group-session/session-auth-group`
- Added `get_auth_group_from_session` function in `GroupSessionController`
- Implemented logic to traverse from `session_id` to `auth_group` through related entities

### Technical Details

- The new endpoint accepts a `session_id `as a query parameter
- The function chain traverses from `GroupSession `-> `Group `-> `Batch `-> `AuthGroup`
- Error handling has been implemented for cases where any entity in the chain is not found

### DIFF
--- a/lib/dbservice_web/controllers/group_session_controller.ex
+++ b/lib/dbservice_web/controllers/group_session_controller.ex
@@ -1,4 +1,7 @@
 defmodule DbserviceWeb.GroupSessionController do
+  alias Dbservice.AuthGroups
+  alias Dbservice.Batches
+  alias Dbservice.Groups
   use DbserviceWeb, :controller
 
   import Ecto.Query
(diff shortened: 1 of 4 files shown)


lib/dbservice_web/controllers/group_session_controller.ex
```diff
@@ -139,4 +142,65 @@ defmodule DbserviceWeb.GroupSessionController do
       |> render("show.json", group_session: group_session)
     end
   end
+
+  swagger_path :get_auth_group_from_session do
+    get("/api/group-session/session-auth-group")
+    description("Fetches the associated auth group data for a session")
+
+    parameters do
+      session_id(:query, :integer, "The ID of the session", required: true)
```
suryabulusu, line 151: why will it be integer.. we usually get string session id

lib/dbservice_web/controllers/group_session_controller.ex
```diff
@@ -139,4 +142,65 @@ defmodule DbserviceWeb.GroupSessionController do
...
+  end
+
+  defp get_group_session(session_id) do
+    case GroupSessions.get_group_session_by_session_id(session_id) do
+      nil -> {:error, :not_found, "Group session not found"}
+      group_session -> {:ok, group_session}
+    end
+  end
+
+  defp get_group(group_id) do
```
suryabulusu, line 186: are we always sure this will be a group

suryabulusu: as in the type being batch. maybe we should be explicit about it. because in future the group_session table may have other groupings (non-batch) too. 

i think aman has written a code somewhere (probably to fetch all sessions corresponding to a group) where he picks rows with type = batch

lib/dbservice_web/controllers/group_session_controller.ex
suryabulusu: ah it seems it will be an integer only.. just checked frontend code. we send ID instead of session_id (from session table)

suryabulusu (commented):
left comments. did you test locally

Bahugunajii (approved):

PR #206, opened 2024-07-30
Added functionality to create user for existing school if user doesn't exist

Bahugunajii:
### Description
This PR introduces functionality to handle scenarios where an existing school does not have an associated user. If a school is found without a user, a new user will be created and associated with that school before updating the school details.

- Updated `update_existing_school_with_user` to check if `existing_school.user_id` is nil.
- If `user_id` is nil, a new user is created using `Users.create_user`
- The new user is then associated with the school, and the school details are updated.
- If `user_id` is not nil, the existing user is retrieved and the school details are updated accordingly.


pritamps (approved):
LGTM!

PR #209, opened 2024-08-07
verify-student endpoint in student controller

Abhinav-3009:
### Overview
Created a new API endpoint for verifying student information. The endpoint allows for the validation of student and user data and group-user record against provided `student_id ` and `auth_group_id ` and other parameters.

### Functionality
1. Accepts a student ID and verification parameters.
2. Retrieves the student and associated user data.
3. Verifies the provided parameters against the student and user data.
4. Retrieves the group-user record using auth_group_id.
5. Returns response as true if verification parameters matches with data and group-user record exists else return false.

### DIFF
--- a/lib/dbservice/groups.ex
+++ b/lib/dbservice/groups.ex
@@ -46,7 +46,7 @@ defmodule Dbservice.Groups do
 
   @doc """
   Gets group by child_id and type.
-  Raises `Ecto.NoResultsError` if the Group does not exist.
+  Raises `Ecto.NoResultsError` if the Group does not exist or if type is wrong/ does not exist.
   ## Examples
       iex> get_group_by_child_id_and_type(123,"batch")
       %Group{}
--- a/lib/dbservice/groups.ex
+++ b/lib/dbservice/groups.ex
@@ -58,6 +58,19 @@ defmodule Dbservice.Groups do
   end
 
   @doc """
+  Gets group by child_id and type.
+  Raises `Ecto.NoResultsError` if the Group does not exist.
+  ## Examples
+      iex> get_group_by_child_id_and_type(123,"batch")
+      %Group{}
+      iex> get_group_by_child_id_and_type(456,"batch")
+      ** (Ecto.NoResultsError)
+  """
+  def get_group_by_child_id_and_type(child_id, type) do
+    Repo.get_by(Group, child_id: child_id, type: type)
+  end
+
+  @doc """
   Creates a group.
   ## Examples
       iex> create_group(%{field: value})
--- a/lib/dbservice/groups.ex
+++ b/lib/dbservice/groups.ex
@@ -45,19 +45,6 @@ defmodule Dbservice.Groups do
   end
 
   @doc """
-  Gets group by child_id.
-  Raises `Ecto.NoResultsError` if the Group does not exist.
-  ## Examples
-      iex> get_group_by_child_id(123)
-      %Group{}
-      iex> get_group_by_child_id(456)
-      ** (Ecto.NoResultsError)
-  """
-  def get_group_by_child_id(child_id) do
-    Repo.get_by(Group, child_id: child_id, type: "batch")
-  end
-
-  @doc """
   Gets group by child_id and type.
   Raises `Ecto.NoResultsError` if the Group does not exist.
   ## Examples
--- a/lib/dbservice_web/controllers/student_controller.ex
+++ b/lib/dbservice_web/controllers/student_controller.ex
@@ -579,7 +579,9 @@ defmodule DbserviceWeb.StudentController do
     post("/api/student/verify-student")
 
     parameters do
-      body(:body, Schema.ref(:VerifyStudentRequest), "Parameters needed to verify student", required: true)
+      body(:body, Schema.ref(:VerifyStudentRequest), "Parameters needed to verify student",
+        required: true
+      )
     end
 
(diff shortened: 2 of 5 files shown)


lib/dbservice/groups.ex
```diff
@@ -57,6 +57,19 @@ defmodule Dbservice.Groups do
...
+  ## Examples
+      iex> get_group_by_child_id_and_type(123,"batch")
+      %Group{}
+      iex> get_group_by_child_id_and_type(456,"batch")
+      ** (Ecto.NoResultsError)
+  """
+  def get_group_by_child_id_and_type(child_id, type) do
+    Repo.get_by(Group, child_id: child_id, type: type)
+  end
+
```
Bahugunajii, line 72: This function is already defined above where type is hardcoded. Maybe we can reuse that and also pass "batch" as argument where it was used.

lib/dbservice_web/controllers/student_controller.ex
```diff
@@ -562,4 +573,89 @@ defmodule DbserviceWeb.StudentController do
...
+  defp verify_student_and_user_data(student, user, verification_params) do
+    Enum.all?(verification_params, fn {key, value} ->
+      cond do
+        key == "auth_group_id" ->
+          verify_auth_group(user.id, value)
+
+        key == "date_of_birth" ->
+          user_dob = Map.get(user, String.to_existing_atom(key))
+          parsed_value = Date.from_iso8601!(value)
+          Date.compare(user_dob, parsed_value) == :eq
```
Bahugunajii, line 632: What happens if `user_dob` and `parsed_value` are not equal?

Bahugunajii (changes requested):

lib/dbservice/groups.ex
Abhinav-3009: Ok, I have done the same in the latest commit.

lib/dbservice_web/controllers/student_controller.ex
Abhinav-3009: It returns false if user_dob and parsed_value are equal, and if any condition is false then the function inside Enum.all() will return false.

lib/dbservice/groups.ex
```diff
@@ -45,16 +45,16 @@ defmodule Dbservice.Groups do
   end
 
   @doc """
-  Gets group by child_id.
+  Gets group by child_id and type.
   Raises `Ecto.NoResultsError` if the Group does not exist.
```
pritamps, line 49: Update comment also to say that error will be returned if type is wrong or does not exist

lib/dbservice_web/controllers/student_controller.ex
```diff
@@ -562,4 +573,89 @@ defmodule DbserviceWeb.StudentController do
       _ -> true
     end
   end
+
+  # function to verify student data with the values recieved in the verification params
+  swagger_path :verify_student do
+    post("/api/student/verify-student")
+
+    parameters do
+      body(:body, Schema.ref(:VerifyStudentRequest), "The verify student request", required: true)
```
pritamps, line 582: Update the message to read bette. Instead of "The verify student requrest", "parameters needed for student veritication" or something?

lib/dbservice_web/swagger_schemas/student.ex
```diff
@@ -209,4 +209,74 @@ defmodule DbserviceWeb.SwaggerSchema.Student do
...
+            student_id(:string, "Student ID")
+            verification_params(Schema.ref(:VerificationParams))
+          end
+
+          required([:student_id, :verification_params])
+
+          example(%{
+            student_id: "20190240808",
+            verification_params: %{
+              auth_group_id: 2,
```
pritamps, line 275: So we can't verify with auth group name (`PunjabStudents`, `DelhiStudents`, etc?)

lib/dbservice_web/controllers/student_controller.ex
```diff
@@ -562,4 +573,89 @@ defmodule DbserviceWeb.StudentController do
...
+
+      {:error, :not_found} ->
+        conn
+        |> put_status(:not_found)
+        |> json(%{error: "Student not found"})
+    end
+  end
+
+  defp get_student_and_user(student_id) do
+    case Users.get_student_by_student_id(student_id) do
```
pritamps, line 608: Why is it `Users.get_student...` and not `Students.get_student`?

pritamps (changes requested):

lib/dbservice_web/swagger_schemas/student.ex
Abhinav-3009: We are getting auth_group_id as query param in the portal-backend from portal-frontend.

lib/dbservice_web/controllers/student_controller.ex
Abhinav-3009: All the Ecto query functions for User, Student and Teacher are present in this Users module.

lib/dbservice_web/controllers/student_controller.ex
Abhinav-3009: Ok, I will update the message.

lib/dbservice/groups.ex
Abhinav-3009: Ok, I will update the comment.

Bahugunajii (approved):

PR #210, opened 2024-08-08
Feat: Enhance GroupUser Update Function for Type-Specific Handling

Bahugunajii:
### Description
- This PR introduces an update function in the GroupUserController to provide type-specific handling when updating group_user and enrollment_record entries.
- The function now accepts a type parameter from the request, which is used to fetch and update the appropriate GroupUser and EnrollmentRecord entries.

### Checklist
- [x] Local testing

### DIFF
--- a/lib/dbservice_web/controllers/group_user_controller.ex
+++ b/lib/dbservice_web/controllers/group_user_controller.ex
@@ -118,20 +118,28 @@ defmodule DbserviceWeb.GroupUserController do
 
     # Fetch the EnrollmentRecord with the specified user_id and where the group_type matches the provided type
     enrollment_record =
-      from(er in EnrollmentRecord,
-        where: er.user_id == ^user_id and er.group_type == ^type
-      )
+      from(er in EnrollmentRecord, where: er.user_id == ^user_id and er.group_type == ^type)
       |> Repo.one()
 
-    # Update both the GroupUser and the EnrollmentRecord, if the enrollment record exists
-    with {:ok, %GroupUser{} = updated_group_user} <-
-           GroupUsers.update_group_user(group_user, params),
-         enrollment_record when not is_nil(enrollment_record) <- enrollment_record,
-         {:ok, %EnrollmentRecord{} = _updated_enrollment_record} <-
-           EnrollmentRecords.update_enrollment_record(enrollment_record, %{
-             "group_id" => new_group_id
-           }) do
-      render(conn, "show.json", group_user: updated_group_user)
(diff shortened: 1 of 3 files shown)


lib/dbservice_web/controllers/group_user_controller.ex
```diff
@@ -105,6 +105,35 @@ defmodule DbserviceWeb.GroupUserController do
...
+    group_user =
+      GroupUsers.get_group_user_by_user_id_and_type(user_id, type)
+      |> List.first()
+
+    # Fetch the EnrollmentRecord with the specified user_id and where the group_type matches the provided type
+    enrollment_record =
+      from(er in EnrollmentRecord,
+        where: er.user_id == ^user_id and er.group_type == ^type
+      )
+      |> Repo.one()
```
pritamps, line 124: Why only one? If it's batch, student can belong to multiple batches no?

lib/dbservice_web/controllers/group_user_controller.ex
```diff
@@ -105,6 +105,35 @@ defmodule DbserviceWeb.GroupUserController do
...
+
+    # Update both the GroupUser and the EnrollmentRecord
+    with {:ok, %GroupUser{} = updated_group_user} <-
+           GroupUsers.update_group_user(group_user, params),
+         {:ok, %EnrollmentRecord{} = _updated_enrollment_record} <-
+           EnrollmentRecords.update_enrollment_record(enrollment_record, %{
```
pritamps, line 130: Shouldn't you check if `enrollment_record` is not `null` or something?

pritamps (changes requested):

lib/dbservice_web/controllers/group_user_controller.ex
Bahugunajii: Added!

lib/dbservice_web/controllers/group_user_controller.ex
pritamps: Hmm okay. Just add a comment to the method that this method assumes that only one ER will be updated so that we don’t anticipate need for multiple updates in same call — just so we don’t lose track of this.

Also add nice doc waala comment to the method :)

pritamps (approved):
Just add one comment and then you can merge

lib/dbservice_web/controllers/group_user_controller.ex
Bahugunajii: Okay!

PR #212, opened 2024-08-14
Feat: Update updated_at for Users on Group Modification

Bahugunajii:
### Description
- Ensure that whenever a group (e.g., batch, school, or auth_group) is updated, the updated_at timestamp of all associated users is also updated.

### Checklist
- [x] Local testing


lib/dbservice/utils/util.ex
```diff
@@ -70,4 +74,30 @@ defmodule Dbservice.Utils.Util do
       end
     end)
   end
+
+  def update_users_for_group(school_id, type) do
```
suryabulusu, line 78: not always school_id

pritamps (approved):
LGTM!

PR #214, opened 2024-09-06
Feat: Add generic update_user_enrollment_records function to StudentController

Bahugunajii:
### Description
This PR introduces a new function update_user_enrollment_records in the StudentController to allow flexible updating of multiple fields in enrollment records based on provided parameters.

### Changes:

- Added a new function update_user_enrollment_records to the StudentController.
- Implemented logic to process any updateable fields passed in the params for enrollment records.
- Created a mechanism to filter out non-updateable fields (student_id and academic_year) from the update operation.
- Utilized the existing EnrollmentRecordView for consistent JSON rendering of updated records.

### DIFF
--- a/lib/dbservice/enrollment_records.ex
+++ b/lib/dbservice/enrollment_records.ex
@@ -143,4 +143,14 @@ defmodule Dbservice.EnrollmentRecords do
 
     Repo.all(query)
   end
+
+  @doc """
+  Fetches all enrollment records for a given user in a specific academic year.
+  """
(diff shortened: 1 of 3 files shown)


pritamps (approved):
Looks good. But what is the use case here?

Bahugunajii:
> Looks good. But what is the use case here?
We recently received a data request indicating that some users were assigned incorrect grades, which has resulted in incorrect grade mapping in their enrollment records. This PR will address the issue by updating the grade fields in the enrollment records for the specified academic year.

PR #222, opened 2024-10-07
Feat: Added functionality to update ER

Bahugunajii:
### Description
- Some students have been incorrectly marked as "dropouts" in our system. We have implemented a process to reverse this mistake by removing the dropout status from both the enrollment records and the student table.

### Checklist
- [x] Self-review


suryabulusu (approved):
looks good

PR #241, opened 2024-12-06
Feat: Remove academic_year for auth_group enrollment records

Bahugunajii:
### Description

This PR addresses an issue with academic year handling for auth group enrollment records by:

- Removing academic_year for auth group during batch processing
- Updating the database to clear existing academic year values for auth groups
- Adding logic to prevent setting academic year for auth group enrollments

### Checklist
- [x] Local testing

### DIFF
--- a/lib/dbservice_web/controllers/group_user_controller.ex
+++ b/lib/dbservice_web/controllers/group_user_controller.ex
@@ -231,12 +231,7 @@ defmodule DbserviceWeb.GroupUserController do
   defp create_new_group_user(conn, params) do
     group = Groups.get_group!(params["group_id"])
 
-    academic_year =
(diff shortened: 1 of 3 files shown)


lib/dbservice_web/controllers/group_user_controller.ex
```diff
@@ -363,12 +370,19 @@ defmodule DbserviceWeb.GroupUserController do
   defp create_new_group_user_from_batch(params) do
     group = Groups.get_group!(params["group_id"])
 
+    academic_year =
+      if group.type == "auth_group" do
+        nil
+      else
+        params["academic_year"]
+      end
+
```
Bahugunajii, line 379: There seems to be repetition of this. Let's avoid repetition.

PR #244, opened 2025-01-04
Feat: Treat grade as a group

Bahugunajii:
### Description
- Added migration to fetch grade information for all users and make an ER & group_user record for each user

### Checklist
- [x] Local testing

### DIFF
--- a/lib/dbservice/grades.ex
+++ b/lib/dbservice/grades.ex
@@ -8,6 +8,7 @@ defmodule Dbservice.Grades do
   alias Dbservice.Repo
 
   alias Dbservice.Grades.Grade
+  alias Dbservice.Groups.Group
 
   @doc """
   Returns the list of grade.
@@ -59,6 +60,7 @@ defmodule Dbservice.Grades do
   def create_grade(attrs \\ %{}) do
     %Grade{}
     |> Grade.changeset(attrs)
+    |> Ecto.Changeset.put_assoc(:group, [%Group{type: "grade", child_id: attrs["id"]}])
     |> Repo.insert()
   end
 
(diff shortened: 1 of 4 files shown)


lib/dbservice/grades.ex
```diff
@@ -59,6 +60,7 @@ defmodule Dbservice.Grades do
   def create_grade(attrs \\ %{}) do
     %Grade{}
     |> Grade.changeset(attrs)
+    |> Ecto.Changeset.put_assoc(:group, [%Group{type: "grade", child_id: attrs["id"]}])
```
suryabulusu, line 63: are u sure these are the only changes? what about group_view.ex ... that file has a rendering for `auth_group` .. shouldn't we have similar code for `grade`? assuming we treat `grade` as group everywhere

priv/repo/migrations/20250104114947_add_grade_to_enrollment_record.exs
```diff
@@ -0,0 +1,85 @@
...
+    SELECT
+        user_id,
+        group_id,
+        group_type,
+        academic_year,
+        start_date,
+        is_current,
+        NOW(),
+        NOW()
+    FROM grade_enrollments;
```
suryabulusu, line 56: shouldn't we be dropping grade_id from ER table?

suryabulusu (commented):
left comments

lib/dbservice/grades.ex
Bahugunajii: It's upto us based on our requirement whether we want to return grade details along with group or not. Added it for now in case we might require it somewhere.

priv/repo/migrations/20250104114947_add_grade_to_enrollment_record.exs
Bahugunajii: Absolutely, we should. However, for now, I'm keeping it as a precautionary measure until we can confirm that everything is working correctly. Will be creating a separate PR to drop it afterwards.

suryabulusu (approved):

PR #245, opened 2025-01-04
Feat: API endpoint to remove specific batch mapping for a student

Bahugunajii:
### Description
- This PR adds functionality to remove a specific batch mapping for a student. This is needed when a student is mapped to incorrect batches and we want to remove one of those mappings.

### Changes:
- Added new DELETE endpoint `/remove-batch/:student_id/:batch_id`
- Added batch lookup using batch_id (string)
- Implemented deletion of corresponding records from:
  - group_user table (using group mapping)
  - enrollment_record table
- Added transaction support to ensure data consistency
- Added proper error handling for various failure cases

The endpoint requires:
- student_id: Student's unique identifier
- batch_id: Batch's string identifier

Error handling covers:
- Student not found
- Batch not found 
- Missing user mapping
- Group mapping not found

Testing:
- [x] Local testing

### DIFF
--- a/lib/dbservice_web/controllers/user_session_controller.ex
+++ b/lib/dbservice_web/controllers/user_session_controller.ex
@@ -9,6 +9,8 @@ defmodule DbserviceWeb.UserSessionController do
   alias Dbservice.Users.Student
   alias Dbservice.Groups.GroupUser
   alias Dbservice.EnrollmentRecords.EnrollmentRecord
+  alias Dbservice.Batches.Batch
+  alias Dbservice.Groups.Group
 
   action_fallback DbserviceWeb.FallbackController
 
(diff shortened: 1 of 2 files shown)


lib/dbservice_web/controllers/user_session_controller.ex
```diff
@@ -207,4 +233,57 @@ defmodule DbserviceWeb.UserSessionController do
...
+          from(gu in GroupUser,
+            where: gu.user_id == ^user_id and gu.group_id == ^group_id
+          )
+          |> Repo.delete_all()
+
+        {:ok, count}
+    end
+  end
+
+  defp delete_batch_enrollment_record(user_id, batch_id) do
```
suryabulusu, line 277: shouldn't we instead set is current to false and an end date?

suryabulusu (approved):
approved

PR #251, opened 2025-01-10
Feat: Updated required fields for school

Bahugunajii:
### Description
- Added af_school_category, district_code, district, state_code and state in the list of required fields

### Checklist
- [x] Local testing


suryabulusu (approved):
approving with caution.. some of these fields are NULL in the table right now

PR #282, opened 2025-04-30
Fix: Funtionality to update user's enrollment records

Bahugunajii:
### Description
- Allow updates to any group_type
- Apply updates only to the fields explicitly passed in the params
- Conditionally update the status record (with group_type = "status" and is_current = true) only if the group_type is "batch"

### Reason for Change
Ops shared data to update batches for some students, but they provided the wrong academic_year. This change allows us to update only specific fields (like academic_year) for the enrollment records using student_id and group_id, without affecting other data.

### Checklist
- [x] Local testing

### DIFF
--- a/lib/dbservice_web/controllers/student_controller.ex
+++ b/lib/dbservice_web/controllers/student_controller.ex
@@ -735,44 +735,81 @@ defmodule DbserviceWeb.StudentController do
   end
 
   def update_user_enrollment_records(conn, params) do
-    student = Users.get_student_by_student_id(params["student_id"])
-    user_id = student.user_id
-    academic_year = params["academic_year"]
+    with student when not is_nil(student) <-
+           Users.get_student_by_student_id(params["student_id"]),
+         user_id <- student.user_id,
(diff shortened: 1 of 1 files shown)


lib/dbservice_web/controllers/student_controller.ex
```diff
@@ -735,44 +735,74 @@ defmodule DbserviceWeb.StudentController do
   end
 
   def update_user_enrollment_records(conn, params) do
-    student = Users.get_student_by_student_id(params["student_id"])
-    user_id = student.user_id
-    academic_year = params["academic_year"]
+    with student when not is_nil(student) <-
```
pritamps, line 738: Are tehre some API docs for db-service? We should add that ASAP if not

pritamps (approved):
Looks good!

lib/dbservice_web/controllers/student_controller.ex
Bahugunajii: We currently have a Swagger doc for db-service, which serves as the API documentation.

PR #297, opened 2025-07-01
 Implemented college model with basic Apis

saltykheera:
image

### DIFF
--- a/lib/dbservice/exams/exam.ex
+++ b/lib/dbservice/exams/exam.ex
@@ -4,11 +4,15 @@ defmodule Dbservice.Exams.Exam do
   use Ecto.Schema
   import Ecto.Changeset
 
+  @primary_key {:exam_id, :string, autogenerate: false}
+  @derive {Phoenix.Param, key: :exam_id}
+
   schema "exam" do
     field :name, :string
     field :registration_deadline, :utc_datetime
     field :date, :utc_datetime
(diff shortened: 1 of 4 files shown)


lib/dbservice/colleges/college.ex
```diff
@@ -0,0 +1,106 @@
...
+
+  @required_fields [
+    :college_id,
+    :institute
+  ]
+
+  @optional_fields [
+    :state,
+    :place,
+    :dist_code,
```
suryabulusu, line 18: see `school` table once.. it says district code... 

even `place` sounds a bit weird. see `user` table.. what is it being called

even `co_ed` ... its `gender_type` in school table. 

i just want to maintain consistency

pritam suggested `type` instead of `college_type` .. what are your thoughts on this?

lib/dbservice_web/controllers/college_controller.ex
```diff
@@ -0,0 +1,163 @@
...
+      CollegeSchema.college(),
+      SwaggerSchemaCommon.group_ids()
+    )
+  end
+
+  @doc """
+  Swagger documentation for the index action.
+  """
+  swagger_path :index do
+    get "/api/colleges"
```
suryabulusu, line 29: why calling it `collges` instead of `college`? just to maintain consistency, for schools it is `/school` instead of `/schools` na?

PR #320, opened 2025-07-30
Get active session occurrences

suryabulusu:
Useful for quizzes -- return occurrences where start_time <= NOW() <= end_time


lib/dbservice_web/controllers/session_occurence_controller.ex
```diff
@@ -65,6 +73,9 @@ defmodule DbserviceWeb.SessionOccurrenceController do
           :is_start_time when value == "today" ->
             from(u in acc, where: u.start_time >= ^today_start and u.start_time <= ^today_end)
 
+          :is_start_time when value == "active" ->
+            from(u in acc, where: u.start_time <= ^current_time and u.end_time >= ^current_time)
+
```
Bahugunajii, line 78:
```suggestion
:is_start_time ->
            case value do
              "today" ->
                from(u in acc, where: u.start_time >= ^today_start and u.start_time <= ^today_end)

              "active" ->
                from(u in acc,
                  where: u.start_time <= ^current_time and u.end_time >= ^current_time
                )
            end
```
Or maybe something more simpler

PR #333, opened 2025-08-06
feat: enhance import functionality with stop capability and CSV template downloads

im-vedant:
- Updated ImportWorker to calculate total rows immediately and handle import halting.
- Added a new TemplateController for downloading CSV templates based on import type.
- Enhanced ImportLive.Index and ImportLive.Show to support stopping imports with confirmation modals.
- Introduced UI elements for displaying stopped imports and improved progress tracking.

### DIFF
--- a/lib/dbservice/utils/import_worker.ex
+++ b/lib/dbservice/utils/import_worker.ex
@@ -293,8 +293,6 @@ defmodule Dbservice.DataImport.ImportWorker do
 
   # Generic single record processing
   defp process_single_record(record, index, import_record, record_processor_fn) do
-    :timer.sleep(5000)
-
     try do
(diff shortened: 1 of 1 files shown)


lib/dbservice/utils/import_worker.ex
```diff
@@ -284,6 +293,8 @@ defmodule Dbservice.DataImport.ImportWorker do
 
   # Generic single record processing
   defp process_single_record(record, index, import_record, record_processor_fn) do
+    :timer.sleep(5000)
```
Bahugunajii, line 296: Remove this

Bahugunajii (changes requested):

PR #335, opened 2025-08-13
Feat: Enhance student import validation to accept either student_id or apaar_id

Bahugunajii:
## Summary
This PR enhances the student data import functionality to provide more flexible identification options. Previously, student imports required a `student_id` field, but now they can accept either `student_id` or `apaar_id` (or both), making the system more adaptable to different data sources.

### 1. Enhanced User Lookup Functionality
- **File**: `lib/dbservice/users.ex`
- **Change**: Added `get_student_by_id_or_apaar_id/1` function
- **Purpose**: Allows finding students by either identifier with priority given to `student_id` when both are present

### 2. Updated Import Validation Logic
- **File**: `lib/dbservice/utils/import_worker.ex`
- **Change**: Modified `process_student_record/1` to validate presence of either identifier
- **Purpose**: Ensures at least one identification field is provided before processing student records

### 3. Updated Field Mapping Requirements
- **File**: `lib/dbservice/constants/mappings.ex`
- **Change**: Made `student_id` optional and `student_apaar_id` required for student imports
- **Purpose**: Aligns validation rules with the new flexible identification approach

## Benefits
- **Flexibility**: Supports different data sources that may use different identification schemes
- **Backward Compatibility**: Existing imports with `student_id` continue to work unchanged
- **Data Integrity**: Maintains validation that at least one identifier is present
- **User Experience**: Reduces import failures due to missing `student_id` when `apaar_id` is available

## Breaking Changes
None. This is a backward-compatible enhancement.

### DIFF
--- a/lib/dbservice/users.ex
+++ b/lib/dbservice/users.ex
@@ -178,6 +178,38 @@ defmodule Dbservice.Users do
   end
 
   @doc """
+  Gets a student by either student_id or apaar_id.
+  Returns the first student found with either identifier.
+
+  ## Examples
+
+      iex> get_student_by_id_or_apaar_id(%{"student_id" => "1234", "apaar_id" => nil})
+      %Student{}
+      iex> get_student_by_id_or_apaar_id(%{"student_id" => nil, "apaar_id" => "123456789101"})
+      %Student{}
+      iex> get_student_by_id_or_apaar_id(%{"student_id" => nil, "apaar_id" => nil})
+      nil
+  """
+  def get_student_by_id_or_apaar_id(%{"student_id" => student_id, "apaar_id" => apaar_id}) do
+    cond do
+      student_id && student_id != "" ->
+        get_student_by_student_id(student_id)
+
+      apaar_id && apaar_id != "" ->
+        Repo.get_by(Student, apaar_id: apaar_id)
+
+      true ->
+        nil
+    end
+  end
+
+  def get_student_by_id_or_apaar_id(record) when is_map(record) do
+    student_id = Map.get(record, "student_id")
+    apaar_id = Map.get(record, "apaar_id")
+    get_student_by_id_or_apaar_id(%{"student_id" => student_id, "apaar_id" => apaar_id})
+  end
+
+  @doc """
   Creates a student.
 
   ## Examples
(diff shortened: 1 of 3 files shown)


lib/dbservice/users.ex
```diff
@@ -177,6 +177,38 @@ defmodule Dbservice.Users do
...
+      %Student{}
+      iex> get_student_by_id_or_apaar_id(%{"student_id" => nil, "apaar_id" => "123456789101"})
+      %Student{}
+      iex> get_student_by_id_or_apaar_id(%{"student_id" => nil, "apaar_id" => nil})
+      nil
+  """
+  def get_student_by_id_or_apaar_id(%{"student_id" => student_id, "apaar_id" => apaar_id}) do
+    cond do
+      student_id && student_id != "" ->
+        get_student_by_student_id(student_id)
```
pritamps, line 196: Why are the two ways of getting student by `student_id` or `apaar_id` different? One is a method, and one uses `Repo.get`?

lib/dbservice/users.ex
```diff
@@ -177,6 +177,38 @@ defmodule Dbservice.Users do
...
+
+      true ->
+        nil
+    end
+  end
+
+  def get_student_by_id_or_apaar_id(record) when is_map(record) do
+    student_id = Map.get(record, "student_id")
+    apaar_id = Map.get(record, "apaar_id")
+    get_student_by_id_or_apaar_id(%{"student_id" => student_id, "apaar_id" => apaar_id})
```
pritamps, line 209: I don't understand this function only :D Is this a recursive function? Why?

lib/dbservice/users.ex
Bahugunajii: The `get_student_by_student_id method` also uses `Repo.get`, and since it’s used in multiple places, I just reused it here for consistency. We can refactor it later to handle everything through a single function.

lib/dbservice/users.ex
Bahugunajii: No this is not a recursive function. This function calls a different function with the same name (it's calling the first clause of the function). This is function overloading/pattern matching.

When we already have a properly structured map like %{"student_id" => "123", "apaar_id" => "456"}, first clause handles this directly

When we have a raw record that might have different field names or structure
record = %{
  "student_id" => "123", 
  "apaar_id" => "456",
  "other_field" => "value",
  "..." => "..."
}

Second clause handles this by extracting only needed fields

PR #350, opened 2025-08-31
Feat: Optimized query

Bahugunajii:
### Description
- Optimized query to fetch user sessions

### Checklist
- [x] Local testing
- [x] Load tested staging environment

### DIFF
--- a/lib/dbservice_web/controllers/user_controller.ex
+++ b/lib/dbservice_web/controllers/user_controller.ex
@@ -1,16 +1,13 @@
 defmodule DbserviceWeb.UserController do
-  alias Dbservice.Groups
   use DbserviceWeb, :controller
 
   import Ecto.Query
   alias Dbservice.Repo
   alias Dbservice.Users
   alias Dbservice.Users.User
-  alias Dbservice.GroupUsers
-  alias Dbservice.GroupSessions
-  alias Dbservice.Sessions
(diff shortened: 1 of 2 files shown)


lib/dbservice_web/controllers/user_controller.ex
```diff
@@ -178,61 +175,69 @@ defmodule DbserviceWeb.UserController do
...
+  # Get user's groups with batch information
+  defp get_user_batch_data(user_id) do
+    from(gu in GroupUser,
+      where: gu.user_id == ^user_id,
+      join: class_group in Group,
+      on: class_group.id == gu.group_id and class_group.type == "batch",
+      join: class_batch in Batch,
+      on: class_batch.id == class_group.child_id,
+      left_join: quiz_group in Group,
+      on: quiz_group.child_id == class_batch.parent_id and quiz_group.type == "batch",
```
pritamps, line 210: Why are you calling it quiz group, but it's of type batch? Maybe call it quiz batch only, no to be consistent?

Bahugunajii: I kept the alias as quiz_group because the join is on the Group table, not the Batch table. Even though it’s a group of type "batch", naming it quiz_batch could be misleading since the source table is still Group

PR #458, opened 2026-04-21
feat: Add paragraphs API and paragraph table

Bahugunajii:
## Summary
- Adds new `paragraph` table with `body` as JSONB and `lang_id` (FK to `language`) via migration `20260421120000_create_paragraph_tables.exs`.
- Introduces `Paragraph` schema + `Paragraphs` context and exposes CRUD endpoints at `/api/paragraph` (router + controller + JSON + swagger schema).
- `GET /api/paragraph/:id` returns the paragraph plus related `problem_lang` rows for the same language by joining `problem_lang` → `resource` where `resource.type_params["paragraph_id"] == paragraph.id` and `problem_lang.lang_id == paragraph.lang_id`.
- Makes `ProblemLanguageJSON.render/1` public for reuse in nested responses.

## Notes / non-goals
- No ordering/`position` logic included (problem_langs returned in stable `id` order).
- Paragraph ↔ problems association is not a new table; it’s derived from existing `resource.type_params.paragraph_id`.

## Test plan
- Run `mix ecto.migrate`
- Start server and verify:
  - `POST /api/paragraph` with `body` + `lang_id`
  - `GET /api/paragraph/:id` includes `problem_langs` when matching `resource.type_params.paragraph_id`

### DIFF
--- a/lib/dbservice/resources/paragraph.ex
+++ b/lib/dbservice/resources/paragraph.ex
@@ -1,0 +1,22 @@
+defmodule Dbservice.Resources.Paragraph do
+  @moduledoc false
+
+  use Ecto.Schema
+  import Ecto.Changeset
+
+  alias Dbservice.Languages.Language
+
+  schema "paragraph" do
+    field(:body, {:array, :map})
+    belongs_to(:language, Language, foreign_key: :lang_id)
+
+    timestamps()
+  end
+
+  @doc false
+  def changeset(paragraph, attrs) do
+    paragraph
+    |> cast(attrs, [:body, :lang_id])
+    |> validate_required([:body, :lang_id])
+  end
+end
--- a/lib/dbservice/resources/paragraph.ex
+++ b/lib/dbservice/resources/paragraph.ex
@@ -5,7 +5,7 @@ defmodule Dbservice.Resources.Paragraph do
   import Ecto.Changeset
 
   schema "paragraph" do
-    field(:body, {:array, :map})
+    field(:body, :string)
 
     timestamps()
   end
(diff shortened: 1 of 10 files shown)


priv/repo/migrations/20260421120000_create_paragraph_tables.exs
```diff
@@ -0,0 +1,13 @@
+defmodule Dbservice.Repo.Migrations.CreateParagraphTables do
+  use Ecto.Migration
+
+  def change do
+    create table(:paragraph) do
+      add :body, :jsonb, null: false
+      add :lang_id, references(:language, on_delete: :nothing), null: false
```
ankurcdoshi: I will check with Trivendu if paragraph can exist without corresponding problem. Because if that is not the case, then we can get rid of this lang_id here, as same lang_id would always be there inside associated problem_lang rows.

Bahugunajii: 👍

ankurcdoshi: Checked with Trivendu - Problems will be added along with paragraph. Hence lang_id is not required in paragraph table.

Bahugunajii: Updated!

lib/dbservice/resources/paragraph.ex
```diff
@@ -0,0 +1,19 @@
+defmodule Dbservice.Resources.Paragraph do
+  @moduledoc false
+
+  use Ecto.Schema
+  import Ecto.Changeset
+
+  schema "paragraph" do
+    field(:body, {:array, :map})
```
ankurcdoshi: We can keep it string, because we will have separate rows to store paragraph in different languages

ankurcdoshi (changes requested):

ankurcdoshi (approved):

PR #471, opened 2026-04-28
fix: DELETE /api/resource/:id by deleting associations first

Bahugunajii:
## Summary
- Fixed `DELETE /api/resource/:id` to delete dependent association rows first (`resource_chapter`, `resource_curriculum`, `resource_topic`, `resource_concept`, `problem_lang`) in a transaction, then delete the `resource`.
- Prevents `Ecto.ConstraintError` crashes from FK constraints like `resource_chapter_resource_id_fkey` / `resource_curriculum_resource_id_fkey`.

## Behavior changes
- Deleting a resource now also removes its join-table/child association records automatically.

## Test plan
- Pick a resource that has chapter/curriculum/topic/concept/problem_lang associations.
- `DELETE /api/resource/:id` → expect `204 No Content`.
- Verify the association rows for that `resource_id` are gone (and the resource is deleted).

### DIFF
--- a/lib/dbservice/resources.ex
+++ b/lib/dbservice/resources.ex
@@ -411,20 +411,7 @@ defmodule Dbservice.Resources do
   """
   def delete_resource(%Resource{} = resource) do
     Repo.transaction(fn ->
-      from(rch in Dbservice.Resources.ResourceChapter, where: rch.resource_id == ^resource.id)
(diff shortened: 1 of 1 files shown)


ankurcdoshi:
Tested to be working

ankurcdoshi (approved):
Working as expected

PR #495, opened 2026-05-06
feat: link comprehension problems to paragraphs and expose paragraph_id in problem_lang

Bahugunajii:
## Summary

Implements storage and API behavior for **reading-comprehension** problems: the passage is persisted in the **`paragraph`** table, and each related **`problem_lang`** row stores a **`paragraph_id`** foreign key plus **`meta_data["paragraph_id"]`** for clients that read question JSON from `meta_data`.

## What changed

- Added **`Dbservice.Resources.Paragraph`** schema and **`Dbservice.Paragraphs`** context (CRUD + helpers used by resource flows).
- **Resource create** (`POST /api/resource`): when the problem is treated as comprehension (`type_params.resource_type` or `subtype` = `comprehension`, case-insensitive), requires passage text via `paragraph_body`, `paragraph`, or nested `meta_data` keys (`passage`, `paragraph_body`, `reading_passage`), creates a **paragraph** row, and sets **`problem_lang.paragraph_id`** and merged **`meta_data.paragraph_id`**.
- **Resource update** (problem + `lang_code`): keeps paragraph body / `meta_data.paragraph_id` in sync for comprehension problems; strips transient keys `paragraph_body` / `paragraph` from attrs passed to changeset.
- **Errors**: missing passage on comprehension create returns **422** with a clear message (`:comprehension_error` in transaction).
- **API docs / responses**: optional **`paragraph_body`** in Swagger for problem resources; **`paragraph_id`** included in problem language JSON and problem list rendering where applicable.

## How to test

- **Postman**: `POST /api/resource` with `type: problem`, `type_params.resource_type: comprehension`, `lang_code`, `paragraph_body`, and question `meta_data` — expect **201**; confirm **`paragraph`** row and **`problem_lang`** FK + `meta_data.paragraph_id`.
- **Regression**: non-comprehension problems unchanged; optional `paragraph_id` on insert for non-comprehension paths if provided in params.

## Notes

- Paragraph API routes were already present; this adds the missing **`Paragraph`** module and wires comprehension creation into the resource transaction.
- Archived SQL dump was moved to `SQL Dump/May 2026/` (unrelated to runtime behavior; omit from release notes if this PR only tracks the feature).

### DIFF
--- a/lib/dbservice/paragraphs.ex
+++ b/lib/dbservice/paragraphs.ex
@@ -1,57 +1,386 @@
 defmodule Dbservice.Paragraphs do
   @moduledoc """
-  Paragraphs store instructional text (`body`); linked problems use
-  `problem_lang` rows with `paragraph_id` (language per problem is `problem_lang.lang_id`).
+  The Paragraphs context.
+
+  Paragraph text for reading-comprehension problems; linked from `problem_lang.paragraph_id`
+  and echoed in `problem_lang.meta_data["paragraph_id"]` for API consumers.
   """
 
   import Ecto.Query, warn: false
   alias Dbservice.Repo
-
-  alias Dbservice.Resources.Resource
-  alias Dbservice.Resources.ProblemLanguage
   alias Dbservice.Resources.Paragraph
+  alias Dbservice.Resources.ProblemLanguage
+  alias Dbservice.Resources.Resource
 
-  def list_paragraph do
-    Repo.all(from(p in Paragraph, order_by: [asc: p.id]))
-  end
+  @doc """
+  Gets a single paragraph.
 
+  Raises `Ecto.NoResultsError` if the paragraph does not exist.
+
+  ## Examples
+
+      iex> fetch_paragraph!(1)
+      %Paragraph{}
+
+      iex> fetch_paragraph!(999_999)
+      ** (Ecto.NoResultsError)
+
+  """
   def fetch_paragraph!(id), do: Repo.get!(Paragraph, id)
 
-  def get_paragraph!(id), do: Repo.get!(Paragraph, id)
+  @doc """
+  Gets a paragraph and all `problem_lang` rows that reference it.
 
-  def get_paragraph_with_problem_langs!(id) do
+  Raises `Ecto.NoResultsError` if the paragraph does not exist.
+
+  ## Examples
+
+      iex> get_paragraph_with_problem_lang!(1)
+      %{paragraph: %Paragraph{}, problem_lang: [%ProblemLanguage{}, ...]}
+
+  """
+  def get_paragraph_with_problem_lang!(id) do
     paragraph = Repo.get!(Paragraph, id)
-    problem_langs = list_problem_langs_for_paragraph(id)
-    %{paragraph: paragraph, problem_langs: problem_langs}
+
+    problem_lang =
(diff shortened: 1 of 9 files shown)


lib/dbservice/paragraphs.ex
```diff
@@ -1,57 +1,396 @@
...
+
+      iex> delete_paragraph(paragraph)
+      {:error, %Ecto.Changeset{}}
+
+  """
+  def delete_paragraph(%Paragraph{} = paragraph), do: Repo.delete(paragraph)
+
+  @doc """
+  Returns true when the resource is a problem marked as comprehension:
+  merged `type_params["resource_type"]`, request `subtype`, or the resource `subtype`,
```
ankurcdoshi: Why are we keeping 3 options here? It will always be resource '`subtype`'. Like other possible subtypes - `mcq_single_answer, mcq_multiple_answer, numerical_answer`, this `comprehension` is also `subtype` for resource when its `type="problem"`

lib/dbservice/paragraphs.ex
```diff
@@ -1,57 +1,396 @@
...
   end
+
+  @doc false
+  defp comprehension_label?(_), do: false
+
+  @doc """
+  Resolves reading-passage body for comprehension workflows from mixed request shapes.
+
+  Checks, in order: `paragraph_body`, `paragraph`, then `meta_data["passage"]`,
+  `meta_data["paragraph_body"]`, `meta_data["reading_passage"]`. Non-map params return `nil`.
```
ankurcdoshi: Why do we need so many options here? Any single name is fine. Maybe you can just keep 
1. '`paragraph`' if having separate `paragraph_id` column in `problem_lang` table
2. Else if separate column '`paragraph_id`' is not there in `problem_lang` table, then '`meta_data[paragraph]`' considering `paragraph_id` node inside `problem_lang` table's `meta_data` column

lib/dbservice/paragraphs.ex
```diff
@@ -1,57 +1,396 @@
 defmodule Dbservice.Paragraphs do
   @moduledoc """
-  Paragraphs store instructional text (`body`); linked problems use
-  `problem_lang` rows with `paragraph_id` (language per problem is `problem_lang.lang_id`).
+  The Paragraphs context.
+
+  Paragraph text for reading-comprehension problems; linked from `problem_lang.paragraph_id`
+  and echoed in `problem_lang.meta_data["paragraph_id"]` for API consumers.
```
ankurcdoshi: If adding `problem_lang.paragraph_id` column then we don't need to echo it again in `problem_lang.meta_data["paragraph_id"]`. Either of the 2 will work.

ankurcdoshi (changes requested):

lib/dbservice/paragraphs.ex
ankurcdoshi: Do we still need '(or override `params["subtype"]`)' here?

ankurcdoshi (changes requested):
1 comment pending now

lib/dbservice/paragraphs.ex
Bahugunajii: We’re only using subtype for comprehension (no type_params / resource_type path anymore). The params["subtype"] part is still needed: on PATCH the client can send the new subtype in the body while resource in memory still has the old subtype until the update is applied, so the check is params["subtype"] || resource.subtype — same field, request wins when present.

ankurcdoshi (approved):

PR #496, opened 2026-05-08
feat: Add batch create/update endpoints for problem resources

Bahugunajii:
## Summary

Adds two endpoints to create and update problem resources in bulk, with **partial-success** semantics and the same body shape as the existing single-resource APIs.

- **POST `/api/resources/problems/batch`** — create many problems in one request.
- **PATCH `/api/resources/problems/batch`** — update many problems in one request.

### `POST /api/resources/problems/batch`
- Body: `{ "problems": [ ...payloads matching POST /api/resource ... ] }`
- `type` defaults to `"problem"` per item.
- Each problem is created in **its own DB transaction**; one failing item does **not** roll back the others.
- 200 response:

  ```json
  {
    "created": [ /* full resource JSON */ ],
    "failed":  [ { "index": 1, "errors": { ... } }, { "index": 3, "error": "..." } ]
  }

### DIFF
--- a/lib/dbservice_web/router.ex
+++ b/lib/dbservice_web/router.ex
@@ -11,7 +11,7 @@ defmodule DbserviceWeb.Router do
   end
 
   pipeline :browser do
-    plug :accepts, ["html"]
+    plug(:accepts, ["html"])
     plug(:fetch_session)
     plug(:fetch_live_flash)
     plug(:put_root_layout, {DbserviceWeb.Layouts, :root})
@@ -20,30 +20,30 @@ defmodule DbserviceWeb.Router do
   end
 
   pipeline :dashboard_auth do
-    plug :admin_basic_auth
+    plug(:admin_basic_auth)
   end
 
   scope "/", DbserviceWeb do
-    pipe_through :browser
+    pipe_through(:browser)
 
-    live "/imports", ImportLive.Index
-    live "/imports/new", ImportLive.New
-    live "/imports/:id", ImportLive.Show
+    live("/imports", ImportLive.Index)
+    live("/imports/new", ImportLive.New)
+    live("/imports/:id", ImportLive.Show)
 
     # Add route for CSV template downloads
-    get "/templates/:type/download", TemplateController, :download_csv_template
+    get("/templates/:type/download", TemplateController, :download_csv_template)
   end
 
   # Protected endpoints (basic auth) for dropout, re-enrollment, auth group, product, program, batch imports
   scope "/", DbserviceWeb do
-    pipe_through [:browser, :dashboard_auth]
+    pipe_through([:browser, :dashboard_auth])
 
-    post "/imports/dropout", ImportController, :create_dropout_import
-    post "/imports/re_enrollment", ImportController, :create_re_enrollment_import
-    post "/imports/auth_group", ImportController, :create_auth_group_import
-    post "/imports/product", ImportController, :create_product_import
-    post "/imports/program", ImportController, :create_program_import
-    post "/imports/batch", ImportController, :create_batch_import
+    post("/imports/dropout", ImportController, :create_dropout_import)
+    post("/imports/re_enrollment", ImportController, :create_re_enrollment_import)
+    post("/imports/auth_group", ImportController, :create_auth_group_import)
+    post("/imports/product", ImportController, :create_product_import)
+    post("/imports/program", ImportController, :create_program_import)
+    post("/imports/batch", ImportController, :create_batch_import)
   end
 
   scope "/api", DbserviceWeb do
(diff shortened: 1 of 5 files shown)


lib/dbservice_web/router.ex
```diff
@@ -117,6 +117,8 @@ defmodule DbserviceWeb.Router do
     get("/resources/curriculum", ResourceController, :curriculum_resources)
     post("/resources/move", ResourceController, :move_resources)
     post("/resources/tests-containing-problems", ResourceController, :tests_containing_problems)
+    post("/resources/problems/batch", ResourceController, :create_problems_batch)
```
ankurcdoshi, line 120: It is creating same paragraph twice when I send 2 problems for same paragraph. json body passed is added at bottom. 

I think paragraph should be taken out from array and added at same level as "problems" representing that both problems belong to same paragraph, hence insert paragraph only once. Something like

```
{
    "paragraph": "p1",
    "problems": [
        {...},
        {...}
    ]
}
```

```
{
    "problems": [
        {
            "lang_code": "en",
            "type": "problem",
            "subtype": "comprehension",
            "paragraph": "paragraph A",
            "skill_ids": [
                1,
                3
            ],
            "type_params": {
                "test_ids": []
            },
            "curriculum_id": 40,
            "grade_id": 4,
            "subject_id": 4,
            "topic_id": 612,
            "difficulty_level": "easy",
            "meta_data": {
                "text": "comprehension q1",
                "options": null,
                "answer": [
                    "2"
                ],
                "solutions": [
                    {
                        "type": "text",
                        "value": "solution sample"
                    }
                ]
            },
            "concept_ids": [
                2,
                3
            ],
            "tags": [
                "tag1",
                "tag2"
            ]
        },
        {
            "lang_code": "en",
            "type": "problem",
            "subtype": "comprehension",
            "paragraph": "paragraph A",
            "skill_ids": [
                2
            ],
            "type_params": {
                "test_ids": []
            },
            "curriculum_id": 40,
            "grade_id": 4,
            "subject_id": 4,
            "topic_id": 613,
            "difficulty_level": "medium",
            "meta_data": {
                "text": "comprehension q2",
                "options": null,
                "answer": [
                    "2.2",
                    "2.3"
                ],
                "solutions": [
                    {
                        "type": "text",
                        "value": "solution sample"
                    }
                ]
            },
            "concept_ids": [
                2
            ],
            "tags": [
                "tag1"
            ]
        }
    ]
}
```

ankurcdoshi (changes requested):

lib/dbservice_web/swagger_schemas/resource.ex
```diff
@@ -417,4 +417,87 @@ defmodule DbserviceWeb.SwaggerSchema.Resource do
...
+      ProblemsBatchUpdateRequest:
+        swagger_schema do
+          title("ProblemsBatchUpdateRequest")
+
+          description(
+            "Batch update; each object includes id (resource id) and PATCH fields like PATCH /api/resource"
+          )
+
+          properties do
+            problems(Schema.array(:object), "Problem patches with id")
```
ankurcdoshi, line 487: Need paragraph property also in case user wants to update it.

ankurcdoshi (changes requested):

lib/dbservice_web/swagger_schemas/resource.ex
```diff
@@ -479,10 +479,16 @@ defmodule DbserviceWeb.SwaggerSchema.Resource do
...
+            paragraph(
+              :string,
+              "Optional reading-comprehension passage; updates the shared paragraph linked to every comprehension problem in this batch"
+            )
+
             problems(Schema.array(:object), "Problem patches with id")
           end
 
           example(%{
+            paragraph: "Updated shared reading passage for the comprehension set.",
```
ankurcdoshi, line 491: Need this paragraph field in single problem update api as well. You can do in this branch or separate - it is up to you.

lib/dbservice_web/router.ex
```diff
@@ -117,6 +117,8 @@ defmodule DbserviceWeb.Router do
...
+    patch("/resources/problems/batch", ResourceController, :update_problems_batch)
```
ankurcdoshi, line 121: difficulty_level & problem meta_data update is not working. It is returning success but no change in database, although paragraph & skill_ids updates are working.

ankurcdoshi (changes requested):

lib/dbservice_web/swagger_schemas/resource.ex
Bahugunajii: Will do it in a separate PR

ankurcdoshi (approved):

PR #510, opened 2026-05-20
feat: include_paragraph_siblings flag for GET /api/problems

Bahugunajii:
## Summary
- Adds optional `include_paragraph_siblings=true` query param to `GET /api/problems`.
- When set, after fetching problems for the requested `topic_id`, any comprehension problem in the result pulls in every other problem sharing the same paragraph — even if those siblings live under a different topic. `curriculum_id` and `lang_code` filters still apply, and results are de-duplicated by resource id.
- Without the flag, behaviour is unchanged: only problems mapped to the requested topic are returned.

### Example
With paragraph `Para1` shared by comprehension problems `P1` (in topic `T1`) and `P2` (in topic `T2`):

- `GET /api/problems?topic_id=1&curriculum_id=1&lang_code=en` → `[P1]`
- `GET /api/problems?topic_id=1&curriculum_id=1&lang_code=en&include_paragraph_siblings=true` → `[P1, P2]`

## Test plan
- [ ] Hit `GET /api/problems?topic_id=<T>&curriculum_id=<C>&lang_code=en` without the flag and confirm the response matches main.
- [ ] Hit it with `&include_paragraph_siblings=true` and confirm comprehension siblings from other topics show up in the list (curriculum/lang still respected).
- [ ] Verify that a topic with no comprehension problems returns the same payload regardless of the flag.
- [ ] Verify de-duplication: a problem that already matches the topic isn't returned twice.

### DIFF
--- a/lib/dbservice_web/controllers/resource_controller.ex
+++ b/lib/dbservice_web/controllers/resource_controller.ex
@@ -1054,19 +1054,21 @@ defmodule DbserviceWeb.ResourceController do
     end
   end
 
-  defp problems_query(topic_id, curriculum_id, lang_code) do
+  defp problems_base_query(curriculum_id, lang_code) do
     from(r in Resource,
+      as: :resource,
       join: rt in ResourceTopic,
+      as: :resource_topic,
       on: rt.resource_id == r.id,
       join: t in Topic,
       on: t.id == rt.topic_id,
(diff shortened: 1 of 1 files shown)


ankurcdoshi (approved):
Tested to be working

PR #514, opened 2026-05-28
feat: add lms_student_documents table + API

pritamps:
## Summary

Adds a new `lms_student_documents` table (and accompanying REST API) so the LMS can store documents (e.g. parent research-consent forms, ID proofs, bonafides) tagged to a student. File bytes live in S3; this table holds metadata + an array of page references in JSONB so multi-page documents are atomic.

- `student_id` FK + `document_type` (allowlist enforced in the changeset, not the DB) + `pages jsonb` + `metadata jsonb` (escape hatch for per-type fields) + `uploaded_by` (email) + `deleted_at` (soft delete) + timestamps
- Indexes: `student_id`, `document_type`, `(student_id, document_type)`, and a partial index on the last for active rows only
- No update endpoint — documents are atomic. Replace = create new + soft-delete old.

## Endpoints

| Method | Path | Notes |
|---|---|---|
| GET    | `/api/lms-student-document` | filter by `student_id` and/or `document_type`; excludes soft-deleted |
| POST   | `/api/lms-student-document` | validates `document_type` allowlist + `pages` shape |
| GET    | `/api/lms-student-document/:id` | 404 on soft-deleted |
| DELETE | `/api/lms-student-document/:id` | soft delete via `deleted_at` |

`pages` JSONB shape (enforced in the changeset, not the DB):

```json
[{"s3_key": "...", "page_number": 1, "mime_type": "image/jpeg", "byte_size": 245678}]
```

## Design notes

- **FK on `student_id`** (`references(:student, on_delete: :nothing)`) — diverges from the FK-less pattern on `lms_pm_school_visit_actions.visit_id`, but since documents flow *through* db-service rather than direct-write from LMS, the consistency is worth having.
- **No DB CHECK on `document_type`** — keeping the allowlist in app code so adding a new type doesn't need a migration.
- **`pages` as JSONB array, not a child table** — matches the `data jsonb` pattern on `lms_pm_school_visit_actions`. The unit of fetch is the whole document; per-page queries aren't a use case.
- **Document type allowlist** starts at: `research_consent`, `id_proof`, `bonafide`, `other`.

## Test plan

- [x] `mix compile --warnings-as-errors` — clean
- [x] `mix credo --strict` on new files — clean
- [x] `mix test` on new controller test — 12/12 pass
- [x] `mix test` full suite — 8 pre-existing failures on `main` (in files this PR doesn't touch; see `e8f048d3` which changed an error message in `group_update_processor.ex` without updating the corresponding test). This PR adds 0 new failures.
- [ ] Deploy to staging; LMS-side upload flow will be wired up in a follow-up PR on `af_lms`.

### DIFF
--- a/lib/dbservice/lms_student_documents.ex
+++ b/lib/dbservice/lms_student_documents.ex
@@ -1,0 +1,71 @@
+defmodule Dbservice.LmsStudentDocuments do
+  @moduledoc """
+  The LmsStudentDocuments context.
+
+  Stores documents (e.g. research consent forms, ID proofs) tagged to a student,
+  with files held in S3 and metadata + page references stored here as JSONB.
+  """
+
+  import Ecto.Query, warn: false
+  alias Dbservice.Repo
+
+  alias Dbservice.LmsStudentDocuments.LmsStudentDocument
+
+  @doc """
+  Lists documents, optionally filtered by student_id and/or document_type.
+  Excludes soft-deleted rows.
+  """
+  def list_lms_student_documents(params \\ %{}) do
+    LmsStudentDocument
+    |> where([d], is_nil(d.deleted_at))
+    |> filter_by(params)
+    |> order_by([d], desc: d.inserted_at)
+    |> Repo.all()
+  end
+
+  defp filter_by(query, params) do
+    Enum.reduce(params, query, fn
+      {"student_id", value}, acc -> where(acc, [d], d.student_id == ^value)
+      {:student_id, value}, acc -> where(acc, [d], d.student_id == ^value)
+      {"document_type", value}, acc -> where(acc, [d], d.document_type == ^value)
+      {:document_type, value}, acc -> where(acc, [d], d.document_type == ^value)
+      _, acc -> acc
+    end)
+  end
+
+  @doc """
+  Gets a single document. Returns `nil` if not found or soft-deleted.
+  """
+  def get_lms_student_document(id) do
+    LmsStudentDocument
+    |> where([d], d.id == ^id and is_nil(d.deleted_at))
+    |> Repo.one()
+  end
+
+  @doc """
+  Creates a document.
+  """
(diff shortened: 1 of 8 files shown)


lib/dbservice/lms_student_documents.ex
```diff
@@ -0,0 +1,71 @@
...
+
+  defp filter_by(query, params) do
+    Enum.reduce(params, query, fn
+      {"student_id", value}, acc -> where(acc, [d], d.student_id == ^value)
+      {:student_id, value}, acc -> where(acc, [d], d.student_id == ^value)
+      {"document_type", value}, acc -> where(acc, [d], d.document_type == ^value)
+      {:document_type, value}, acc -> where(acc, [d], d.document_type == ^value)
+      _, acc -> acc
+    end)
+  end
```
Bahugunajii, line 34: Maybe we can use:

```suggestion
  defp filter_by(query, params) do
  Enum.reduce(params, query, fn
    {key, value}, acc ->
      case normalize_key(key) do
        :student_id ->
          where(acc, [d], d.student_id == ^value)

        :document_type ->
          where(acc, [d], d.document_type == ^value)

        _ ->
          acc
      end
  end)
end

defp normalize_key(key) when is_binary(key), do: String.to_existing_atom(key)
defp normalize_key(key), do: key
```

Bahugunajii (changes requested):

lib/dbservice/lms_student_documents.ex
pritamps: Applied — with one small tweak: wrapped `String.to_existing_atom` in a `rescue ArgumentError` so an unknown query key (e.g. a typo like `?student_idd=1`) returns `:unknown` and gets ignored, rather than crashing the route. Commit: 24b15719

Bahugunajii (approved):

PR #530, opened 2026-06-03
feat(gurukul): config resolution endpoint

Bahugunajii:
## What

Adds `GET /api/gurukul-config` which resolves the merged Gurukul UI configuration along the fallback chain `batch → program → defaultgroup` (later layers win).

**Step 2 of 2.** Stacked on #529 (program `config` column) — review/merge that first; this PR's base will retarget to `main` automatically once #529 merges.

## Resolution rules

- **Layers (later wins):** `defaultgroup` (auth_group named `"defaultgroup"`, under `input_schema["gurukul_config"]`) ← `program.config` ← `batch.metadata["gurukul_config"]`
- **For a user:** pick the **oldest current batch** (earliest `start_date`, then `inserted_at`) → merge with its program + defaultgroup. No batch → oldest current program. Neither → defaultgroup. This implements the agreed "older batch wins" rule for students in multiple batches.
- Returns `{config, resolved_from}`, where `resolved_from` records the source layer + batch/program ids for observability.

## API

`GET /api/gurukul-config?user_id=` | `?batch_id=` | `?program_id=`

```json
{"config": {"showTests": true, "testsSectionTitle": "CA Test"},
 "resolved_from": {"source": "batch", "batch_id": 12, "program_id": 4}}
```

## Changes

- `Dbservice.Services.GurukulConfigService` — merge + fallback logic
- `GurukulConfigController` + `GurukulConfigJSON` — endpoint and serializer
- Router — new route
- Tests — `gurukul_config_service_test.exs` (8 tests: precedence, oldest-batch-wins, program fallback, empty defaults)

## Testing

- 8/8 tests pass (SQL sandbox)
- `mix compile --warnings-as-errors`, `mix format --check-formatted`, `mix credo --strict` clean
- HTTP smoke test (with bearer token): 400s for missing/invalid params; 200 with defaultgroup fallback for unknown user/program

## Note on `defaultgroup`

No `defaultgroup` auth_group exists yet — resolution degrades gracefully to an empty base. Seeding it + populating real configs is follow-up work (along with the Gurukul frontend migration to this endpoint).

### DIFF
--- a/lib/dbservice_web/controllers/gurukul_config_controller.ex
+++ b/lib/dbservice_web/controllers/gurukul_config_controller.ex
@@ -46,12 +46,16 @@ defmodule DbserviceWeb.GurukulConfigController do
   defp resolve(:batch, id), do: GurukulConfigService.resolve_for_batch(id)
   defp resolve(:program, id), do: GurukulConfigService.resolve_for_program(id)
 
+  @scope_params [{"user_id", :user}, {"batch_id", :batch}, {"program_id", :program}]
+
   defp pick_scope(params) do
-    cond do
-      present?(params["user_id"]) -> {:ok, :user, params["user_id"]}
-      present?(params["batch_id"]) -> {:ok, :batch, params["batch_id"]}
-      present?(params["program_id"]) -> {:ok, :program, params["program_id"]}
-      true -> {:error, "One of user_id, batch_id or program_id is required"}
+    # The three params are mutually-exclusive ways to resolve config for a
+    # single entity, not filters to combine. Reject ambiguous requests rather
+    # than silently honouring one and ignoring the rest.
+    case Enum.filter(@scope_params, fn {key, _scope} -> present?(params[key]) end) do
+      [{key, scope}] -> {:ok, scope, params[key]}
+      [] -> {:error, "One of user_id, batch_id or program_id is required"}
+      _ -> {:error, "Provide only one of user_id, batch_id or program_id"}
     end
   end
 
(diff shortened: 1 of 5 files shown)


lib/dbservice_web/controllers/gurukul_config_controller.ex
```diff
@@ -0,0 +1,68 @@
...
+    end
+  end
+
+  defp resolve(:user, id), do: GurukulConfigService.resolve_for_user(id)
+  defp resolve(:batch, id), do: GurukulConfigService.resolve_for_batch(id)
+  defp resolve(:program, id), do: GurukulConfigService.resolve_for_program(id)
+
+  defp pick_scope(params) do
+    cond do
+      present?(params["user_id"]) -> {:ok, :user, params["user_id"]}
```
suryabulusu: didn't get this.. should we be blocking if multiple params are sent? or validate all if multiple ones are sent?

example: /api/gurukul-config?user_id=12&batch_id=99 

right now it just checks for user id = 12

doesn't validate if batch id is 99 or not

suryabulusu (commented):
left comment

lib/dbservice_web/controllers/gurukul_config_controller.ex
Bahugunajii: These three params aren't filters to combine, they're mutually-exclusive ways to resolve config for a single entity (resolve for this user OR this batch OR this program), which is why the swagger says "provide exactly one." So neither silent first-wins nor cross-validating (does batch 99 belong to user 12?) is the right behavior the latter would be inventing a constraint the endpoint doesn't need.

I've made pick_scope reject ambiguity instead: more than one param now returns a 400 "Provide only one of user_id, batch_id or program_id", so ?user_id=12&batch_id=99 is a clear error rather than silently resolving user 12. The frontend only ever sends a single user_id, so nothing live is affected. Done in 63755b3.

suryabulusu (approved):

PR #538, opened 2026-06-10
feat(resources): reject duplicate test code on resource create

Bahugunajii:
## Summary
New requirement for the create resource endpoint (`POST /api/resource`): when `type='test'` and the test `code` in the request body already exists in the database, the API now returns an error instead of updating the existing test resource.

## Changes
- `handle_test_resource/2` in `resource_controller.ex` now returns `422 Unprocessable Entity` with `{"error": "This test code has already been used."}` when a resource with the given test code already exists.
- New test codes (no existing match) still create a new resource as before.

## Notes
- If `code` is missing/`nil`, behavior is unchanged (falls through to creation).

### DIFF
--- a/lib/dbservice_web/controllers/resource_controller.ex
+++ b/lib/dbservice_web/controllers/resource_controller.ex
@@ -82,13 +82,8 @@ defmodule DbserviceWeb.ResourceController do
     code = params["code"]
 
     case Resources.get_resource_by_code(code) do
-      nil ->
-        create_new_resource(conn, params)
-
-      _existing_resource ->
-        conn
-        |> put_status(:unprocessable_entity)
-        |> json(%{error: "This test code has already been used."})
+      nil -> create_new_resource(conn, params)
+      _existing_resource -> {:error, "This test code has already been used."}
     end
   end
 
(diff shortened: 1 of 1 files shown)


lib/dbservice_web/controllers/resource_controller.ex
```diff
@@ -82,8 +82,13 @@ defmodule DbserviceWeb.ResourceController do
...
     case Resources.get_resource_by_code(code) do
-      nil -> create_new_resource(conn, params)
-      existing_resource -> update_existing_resource(conn, existing_resource, params)
+      nil ->
+        create_new_resource(conn, params)
+
+      _existing_resource ->
+        conn
+        |> put_status(:unprocessable_entity)
+        |> json(%{error: "This test code has already been used."})
```
ankurcdoshi: Can we have common error pattern?

PATCH api `http://localhost:4000/api/resource/1039` returns following error when 0 is sent for `topic_id`
```
{
    "errors": {
        "message": "Topic not found",
        "detail": "Unprocessable Entity"
    }
}
```

And here it throws error like 
```
{
    "error": "This test code has already been used."
}
```

ankurcdoshi (changes requested):

ankurcdoshi (approved):

PR #543, opened 2026-06-12
Add program_id to centres with type-based backfill

pritamps:
## Summary

- Adds a nullable, indexed `centres.program_id` foreign key referencing `program(id)`.
- Backfills the column from the stable Centre type codes in the same migration: `type_code = 'coe'` → program "JNV CoE", `type_code = 'nodal'` → program "JNV Nodal". Programs are matched **by name**, not id, so the migration is environment-safe across staging and production.
- Extends the Centre schema-contract test to cover the new column (nullability, index, FK boundary).

## Context

Centres were imported to staging and production from the finance cost-centre export (af_lms#114 / #533). Teachers in `user_permission` carry `program_ids`, and program is the natural differentiator where one school hosts multiple centres (currently JNV Adilabad: CoE + Nodal). Putting `program_id` on the centre makes that relationship explicit and prepares for teacher↔centre mapping (deliberately **not** in this PR — teachers need to become full db-service users first).

The af_lms side (program search + selector in the Centre edit modal, card display) follows in a separate LMS PR gated on this schema.

## Verification

- `mix format --check-formatted`
- `mix test test/dbservice/centre_schema_test.exs` (1 test, 0 failures — test DB runs the new migration)

## Deployment notes

- Staging: the gated "Run Staging Migrations" workflow on this PR applies it.
- Production: lands with the next `main` → `release` push.
- Backfill affects 47 of 54 rows in prod (36 coe + 11 nodal by type); centres with other/no types keep `program_id = NULL`.

Refs avantifellows/af_lms#105.

### DIFF
--- a/priv/repo/migrations/20260612090000_add_program_id_to_centres.exs
+++ b/priv/repo/migrations/20260612090000_add_program_id_to_centres.exs
@@ -1,0 +1,38 @@
+defmodule Dbservice.Repo.Migrations.AddProgramIdToCentres do
+  use Ecto.Migration
+
+  def up do
+    alter table(:centres) do
+      add :program_id, references(:program, on_delete: :nothing)
+    end
+
+    create index(:centres, [:program_id])
+
+    # Backfill from the stable Centre type codes. Matching programs by name
+    # keeps the migration environment-safe (program ids are not guaranteed
+    # to match across staging and production).
+    execute("""
+    UPDATE centres
+    SET program_id = program.id
+    FROM program
+    WHERE program.name = 'JNV CoE'
+      AND centres.type_code = 'coe'
+      AND centres.program_id IS NULL
+    """)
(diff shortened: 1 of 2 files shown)


Bahugunajii (approved):

deepansh96:
are we sure that one centre will only ever belong to one program id?

deepansh96:
backfilling have you tried once in staging or local based on the prod data?

pritamps:
> are we sure that one centre will only ever belong to one program id?

Yeah. For now, this is true. `ever` ka pata nahi. We will cross that bridge when we get to it. But for foreseeable future, this is true

pritamps:
> backfilling have you tried once in staging or local based on the prod data?

Not yet! Will do on staging now that PR is approved. But it's quite straightforward

PR #545, opened 2026-06-12
Add staff and centre_positions tables; make teacher_id optional and unique

pritamps:
## Summary

Schema groundwork for managing AF staff (teachers, PMs, APCs) and centre seat assignments from the LMS. Four migrations + one changeset change:

1. **`teacher` changes**
   - `teacher_id` becomes **optional** (changeset-level; the column was already nullable). Teachers backfilled from the LMS may not have a confirmed AF employee ID yet — admins assign codes later via the upcoming Staff Management UI. A code-less teacher simply cannot log in via Portal `/teacher/verify` until the code is set (today they have no row at all, so this is strictly additive).
   - `teacher_id` becomes **unique** (`teacher_teacher_id_unique`, supersedes the plain lookup index from 20250825075810; NULLs exempt). `teacher_id` is the Gurukul/Portal login identifier — admin-entered codes must never collide. The three rows holding the `"Not Permanent"` placeholder are nulled out first (that is what the placeholder meant).
   - New `exit_date` (date, NULL = active), exposed through the existing teacher API.
2. **`staff` (new)** — NON-teaching AF staff (program managers, APCs, future ops roles): unique `employee_code` (AF###), unique `user_id`, `staff_type`, `designation`, `exit_date`. Teachers remain fully modelled by `teacher` (single source for their code = login id; no mirroring/sync). A person may hold rows in both tables (e.g. teacher promoted to APC), same code by app-level convention.
3. **`centre_positions` (new)** — seats at centres: `centre_id`, `role` (physics/chemistry/maths/biology/apc/pm), nullable occupant `user_id` (**NULL = vacant/TBH**), `hr_code` (MT###/TBH### reference into HR systems), notes, soft delete. Partial unique index on active assignments `(centre_id, role, user_id)` — multiple vacancies per role allowed; PMs may hold seats at many centres.
4. **`user_permission`** — explicit `user_id` FK (the email join is fragile: `user.email` is nullable/unverified-unique) and `revoked_at` (access revocation as a soft state instead of today's hard-delete offboarding).

Also reworks the teacher controller invalid-payload fixtures to stay invalid (via the user phone-format validation) now that `teacher_id` is optional, and adds a schema-contract test (`staff_positions_schema_test.exs`) covering all of the above.

## Design context

- Occupancy is keyed on `user_id` (not `teacher_id`) so the same mechanism covers teachers, PMs, and APCs.
- Vacancies in HR data are placeholder "TBH" employees; here they are simply unoccupied position rows.
- `staff` and `centre_positions` are LMS-written tables (like `centres`); no Ecto contexts/APIs are added for them. Teacher writes continue through the existing teacher REST API.
- Full design narrative: `af_lms/docs/ai/teacher-staff/2026-06-12-centres-and-staff-handoff.md`.

## Verification

- `mix format --check-formatted`
- `mix test test/dbservice/staff_positions_schema_test.exs test/dbservice/centre_schema_test.exs test/dbservice/users_test.exs test/dbservice_web/controllers/teacher_controller_test.exs` — 43 tests, 0 failures (fresh test DB)
- Full `mix test`: 389 tests, 8 failures — identical to origin/main baseline (same pre-existing unrelated failures noted on #533/#543)

## Deployment notes

- Staging: gated "Run Staging Migrations" workflow on this PR.
- Production: next `main` → `release` push (can ride together with #543).
- The "Not Permanent" null-out touches 3 rows; the unique index creation will fail loudly if any real duplicate codes exist at migration time (none do today: 3,007 rows, all codes distinct apart from the placeholder).

Refs avantifellows/af_lms#105.

### DIFF
--- a/lib/dbservice/users.ex
+++ b/lib/dbservice/users.ex
@@ -576,13 +576,20 @@ defmodule Dbservice.Users do
 
   @doc """
   Gets a Teacher by teacher ID.
-  Raises `Ecto.NoResultsError` if the Batch does not exist.
+  Returns nil when no teacher matches, or when the given teacher ID is
+  nil/blank (teacher_id is optional, so code-less teachers exist and
+  callers may pass through missing request/import values).
   ## Examples
-      iex> get_teacher_by_teacher_id(1234)
+      iex> get_teacher_by_teacher_id("AF123")
       %Teacher{}
-      iex> get_teacher_by_teacher_id(abc)
-      ** (Ecto.NoResultsError)
+      iex> get_teacher_by_teacher_id("no-such-code")
+      nil
+      iex> get_teacher_by_teacher_id(nil)
+      nil
   """
+  def get_teacher_by_teacher_id(teacher_id) when is_nil(teacher_id) or teacher_id == "",
+    do: nil
+
   def get_teacher_by_teacher_id(teacher_id) do
     Repo.get_by(Teacher, teacher_id: teacher_id)
   end
--- a/lib/dbservice/users/teacher.ex
+++ b/lib/dbservice/users/teacher.ex
@@ -12,6 +12,7 @@ defmodule Dbservice.Users.Teacher do
     field :designation, :string
     field :teacher_id, :string
     field :is_af_teacher, :boolean
+    field :exit_date, :date
 
     belongs_to :user, User
     has_one :teacher_profile, TeacherProfile
@@ -28,8 +29,10 @@ defmodule Dbservice.Users.Teacher do
       :designation,
       :teacher_id,
       :subject_id,
-      :is_af_teacher
+      :is_af_teacher,
+      :exit_date
     ])
-    |> validate_required([:user_id, :teacher_id])
+    |> validate_required([:user_id])
+    |> unique_constraint(:teacher_id, name: :teacher_teacher_id_unique)
   end
 end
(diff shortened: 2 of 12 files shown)


deepansh96:
One concern: this makes teacher_id optional, but a few existing flows still assume it is always present. For example, get_teacher_by_teacher_id/1 crashes if it receives nil, and teacher imports / teacher profile creation can pass nil into it.

Can we guard nil/blank teacher_id in the lookup helper, and make teacher_id optional in the teacher import mapping too? That should keep code-less teacher creation safe.

deepansh96:
Small follow-up: exit_date is added to the teacher schema/changeset, but it is not included in the teacher JSON response yet. Can we add it to TeacherJSON too, so API consumers can read it back after create/update? This will be helpful for the LMS side as well.

pritamps:
> One concern: this makes teacher_id optional, but a few existing flows still assume it is always present. For example, get_teacher_by_teacher_id/1 crashes if it receives nil, and teacher imports / teacher profile creation can pass nil into it.
> 
> Can we guard nil/blank teacher_id in the lookup helper, and make teacher_id optional in the teacher import mapping too? That should keep code-less teacher creation safe.

Not in import (will check, but maybe there too, yeah) - but will fix the `get_teacher_by_techer_id` one

Fixed both, @deepansh96

pritamps:
> Small follow-up: exit_date is added to the teacher schema/changeset, but it is not included in the teacher JSON response yet. Can we add it to TeacherJSON too, so API consumers can read it back after create/update? This will be helpful for the LMS side as well.

Good point - will add!

EDIT: Added!

Bahugunajii (approved):

PR #558, opened 2026-06-18
feat(student): scope student identity to (student_id, auth_group)

Bahugunajii:
## Why

`student.student_id` is **not** globally unique — the same id (often a phone number) is reused across auth groups (e.g. `DelhiStudents`, `BiharStudents`). The POST-student flow matched any student row with that `student_id` and **updated** it, causing cross-auth-group data leakage: one student row shared across auth groups, with batch/grade/school enrollments and `group_user` mappings overwritten.

A student is now identified by **`(student_id, auth_group)`**. The auth_group **enrollment record** (`group_type="auth_group"`, `group_id = auth_group.id`, `is_current=true`) is the source of truth for ownership.

The real registration path is the plain `POST /student` → `create_or_update_student` (used by portal-backend and the CSV import worker); `/student/create-with-enrollments` is unused.

## What

- **`create_or_update_student/1`** is auth-group-scoped and transactional when an `auth_group` is present (new `get_student_by_student_id_and_auth_group/2` + `resolve_auth_group_id/1`). Ownership ER + `group_user` are created for **new** students via the existing `EnrollmentService` (idempotent on group_user). Legacy global behavior is kept as a fallback when no auth group is supplied. **`apaar_id` stays globally unique.**
- **`GET /student`** accepts `auth_group` / `auth_group_id` and scopes results via the auth_group enrollment record join (lets portal-backend scope its verify lookup).
- **Import worker** `process_student_record` scopes the "already exists" check to the auth group and wraps create + enrollments in a single transaction.
- Tests for the three outcomes: same `student_id` across two auth groups → two rows; same `(student_id, auth_group)` → update; duplicate `apaar_id` across groups → rejected.

No DB migration needed (`student.student_id` already has only a non-unique index).

## Verification

- `mix credo` clean, `mix test test/dbservice/users_test.exs` → 35/35.
- Manual end-to-end against a running server: same `student_id` POSTed under `AllIndiaStudents` then `MaharashtraStudents` → two distinct rows, each with its own current auth_group ER; re-POST under the same auth group → update (no duplicate). Test data cleaned up afterward.

## Coordination

Pairs with **portal-backend PR avantifellows/portal-backend#88** (sends `auth_group` in the create body, drops the redundant ownership call, scopes `verify_student_by_id`). Both must deploy together.

### DIFF
--- a/lib/dbservice/users.ex
+++ b/lib/dbservice/users.ex
@@ -7,6 +7,10 @@ defmodule Dbservice.Users do
   alias Dbservice.Utils.Util
   alias Dbservice.Repo
 
+  alias Dbservice.AuthGroups
+  alias Dbservice.EnrollmentRecords.EnrollmentRecord
+  alias Dbservice.Services.EnrollmentService
+
   alias Dbservice.Users.User
 
   @doc """
@@ -226,6 +230,60 @@ defmodule Dbservice.Users do
   end
 
   @doc """
+  Gets a student scoped to an auth group.
+
+  A `student_id` is only unique within the scope of an auth group, so a student is the
+  *same logical student* only when its `student_id` matches AND there is a current
+  enrollment record tying it to the given auth group (`group_type = "auth_group"`,
+  `group_id = auth_group.id`, `is_current = true`).
+
+  Returns the matching `%Student{}` or `nil`.
+  """
+  def get_student_by_student_id_and_auth_group(student_id, auth_group_id)
+      when not is_nil(student_id) and not is_nil(auth_group_id) do
+    from(s in Student,
+      join: er in EnrollmentRecord,
+      on: er.user_id == s.user_id,
+      where:
+        s.student_id == ^student_id and
+          er.group_type == "auth_group" and
+          er.group_id == ^auth_group_id and
+          er.is_current == true,
+      limit: 1
+    )
+    |> Repo.one()
+  end
+
+  def get_student_by_student_id_and_auth_group(_student_id, _auth_group_id), do: nil
+
+  @doc """
+  Resolves the incoming auth group id from request/import params.
+
+  Prefers an explicit `auth_group_id` (the data-import pipeline already sets this to
+  `auth_group.id`); otherwise resolves the `auth_group` name via `AuthGroups`.
+  Returns the `auth_group.id` (which is what enrollment records store as `group_id`),
+  or `nil` when no auth group can be determined.
+  """
+  def resolve_auth_group_id(params) do
+    case params["auth_group_id"] do
+      id when not is_nil(id) and id != "" ->
+        id
+
+      _ ->
+        resolve_auth_group_id_by_name(params["auth_group"])
+    end
+  end
+
+  defp resolve_auth_group_id_by_name(name) when is_binary(name) and name != "" do
+    case AuthGroups.get_auth_group_by_name(name) do
+      nil -> nil
+      auth_group -> auth_group.id
+    end
+  end
+
+  defp resolve_auth_group_id_by_name(_name), do: nil
+
+  @doc """
   Creates a student.
 
   ## Examples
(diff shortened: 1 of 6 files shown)


lib/dbservice/users.ex
suryabulusu: ** (Ecto.MultipleResultsError) expected at most one result but got 2

this function is used by update_student_status i think. and u will get the above error when two rows have same student id.


 get_student_by_id_or_apaar_id  -- this function will create issue too... but i think it will return something (potentially incorrect) instead of failing.


need to correct these two or not use these two functions

suryabulusu (commented):
left a comment.. i think important enough to correct

lib/dbservice/users.ex
```diff
@@ -352,22 +432,100 @@ defmodule Dbservice.Users do
...
+        case upsert_student_scoped(existing_student, params, auth_group_id) do
+          {:ok, student} -> student
+          {:error, reason} -> Repo.rollback(reason)
+        end
+      end)
+    end
+  end
+
+  # New student: create it and establish auth-group ownership (enrollment record + group_user).
+  defp upsert_student_scoped(nil, params, auth_group_id) do
```
suryabulusu, line 487: I think this can silently create a **second student row with the same `student_id` in the same auth group**, which is exactly what we're trying to prevent.

The scoped lookup (`get_student_by_student_id_and_auth_group`) only matches when there's an auth_group enrollment record with `is_current = true`. So if a student's auth_group ER is missing or got flipped to `is_current = false` (which happens in our dropout / re-enrollment flow), the lookup returns `nil`, we land in this `upsert_student_scoped(nil, ...)` clause, and `create_student_with_user` inserts a fresh row.

The only validation on this path is `validate_apaar_id_not_duplicated` (apaar only) — there's no `student_id` duplicate check here, and there's no DB unique constraint on `student.student_id` (only a non-unique index), so nothing else catches it.

Example:
1. Student `S1` exists in `DelhiStudents`, gets dropped → its auth_group ER goes `is_current = false`.
2. Portal re-POSTs `S1` under `DelhiStudents`.
3. Scoped lookup finds no *current* ER → `nil` → we create a duplicate `S1` row in `DelhiStudents`.

Suggest a partial unique index on `(student_id, auth_group ownership)` as a real backstop, or at minimum a `student_id`-within-auth-group check before creating.

lib/dbservice/users.ex
```diff
@@ -352,22 +432,100 @@ defmodule Dbservice.Users do
...
+  end
+
+  defp auth_group_name_from_id(auth_group_id) do
+    case AuthGroups.get_auth_group!(auth_group_id) do
+      %{name: name} -> name
+      _ -> nil
+    end
+  end
+
+  defp validate_apaar_id_not_duplicated(apaar_id, existing_student) do
```
suryabulusu, line 555: Separate issue, on the helper this calls (`duplicate_exists?`, ~line 687): it uses `Repo.one`, which **raises `Ecto.MultipleResultsError` when more than one row matches**. `validate_identifiers_not_duplicated` on the global path calls it with `:student_id`.

Given the whole premise of this PR is that the same `student_id` legitimately exists across auth groups, the moment a `student_id` is shared by 2+ other rows, that line raises instead of returning a clean error → the create/update 500s.

Example: two students already share `student_id = "9876543210"`. A POST without `auth_group` (falls back to global) hits `duplicate_exists?(:student_id, "9876543210", ...)` → `Repo.one` sees 2 rows → crash.

`Repo.exists?` (or `Repo.aggregate(:count)`) instead of `Repo.one` would be safe against the duplicates we now expect.

lib/dbservice/users.ex
```diff
@@ -165,16 +169,38 @@ defmodule Dbservice.Users do
...
-  def get_student_by_student_id(student_id) do
-    Repo.get_by(Student, student_id: student_id)
+  def get_student_by_student_id(student_id, auth_group_params \\ %{}) do
+    case resolve_auth_group_id(auth_group_params) do
+      nil -> get_student_by_student_id_unscoped(student_id)
+      auth_group_id -> get_student_by_student_id_and_auth_group(student_id, auth_group_id)
+    end
+  end
+
+  defp get_student_by_student_id_unscoped(student_id) do
```
suryabulusu, line 197: Heads up on a quiet behavior change for callers that **don't** pass an auth_group. The old `get_student_by_student_id` was `Repo.get_by(...)`, which *raises* `Ecto.MultipleResultsError` on duplicate `student_id`s. This new unscoped version does `order_by: [asc: id], limit: 1` and silently returns the earliest row.

That trade-off is fine where it's intentional, but the unscoped (arity-1) callers include identity-sensitive ones that weren't updated to pass auth_group: `user_session_controller` login (lines 140/165), `student_profile_controller`, and `student_update_service`. With duplicate `student_id`s now expected, those can silently resolve to the *wrong* student (e.g. authenticate against whoever was created first) with no error.

Not necessarily blocking, but we should either pass auth_group through on those paths or be explicit that login/profile-by-student_id is now ambiguous.

suryabulusu (commented):
Took a careful pass focused on the new auth-group-scoped paths. Overall the approach is right, but I found a few gaps where the duplicate-protection / identity-validation guarantees don't fully hold. The first four fire on the exact path portal-backend #88 now uses (POST /student with auth_group), so they're worth fixing before we deploy the pair. The last one is more of a future footgun. Inline below.

lib/dbservice/users.ex
```diff
@@ -352,22 +432,100 @@ defmodule Dbservice.Users do
...
+  defp upsert_student_scoped(nil, params, auth_group_id) do
+    with {:ok, student} <- create_student_with_user(params),
+         {:ok, _ownership} <- ensure_auth_group_ownership(student, params, auth_group_id) do
+      {:ok, student}
+    end
+  end
+
+  # Existing student in this auth group: update fields only. Ownership already exists (it is
+  # what the scoped lookup matched on), so it must not be re-created.
+  defp upsert_student_scoped(%Student{} = existing_student, params, _auth_group_id) do
```
suryabulusu: This update branch calls `update_student_with_user` directly, which skips the identifier-mismatch check the global path runs (`validate_identifier_matches` → `validate_apaar_id_match`). So on the scoped update path a request can silently **overwrite** the stored `apaar_id`.

Example: existing student `(student_id=S1, auth_group=G, apaar_id=A_old)`. Portal re-POSTs `(S1, G, apaar_id=A_new)` where `A_new` isn't used by anyone else. We match by `(S1, G)`, `validate_apaar_id_not_duplicated` passes (nobody else owns `A_new`), and we silently change apaar to `A_new`. The global path would have rejected this with the "APAAR ID doesn't match existing" error.

Since portal re-POSTs the same student to update fields, this path is live. Worth re-running the apaar-matches-existing check before updating.

lib/dbservice_web/controllers/student_controller.ex
```diff
@@ -236,8 +271,9 @@ defmodule DbserviceWeb.StudentController do
   end
 
   def enrolled(conn, params) do
-    # Retrieve the student information based on the provided student ID
-    student = Users.get_student_by_student_id(params["student_id"])
+    # Retrieve the student information based on the provided student ID, scoped to the auth
+    # group when one is supplied (student_id is unique only within an auth group).
+    student = Users.get_student_by_student_id(params["student_id"], params)
     user_id = student.user_id
```
suryabulusu, line 288: Minor / future footgun, not reachable via portal today: `student` here can be `nil` and `student.user_id` would 500.

With the lookup now taking `params`, if a caller ever includes `auth_group`/`auth_group_id` in the `/enrolled` payload (portal #88 only adds it to POST /student and the verify lookup, not here), a student that previously matched by `student_id` alone could return `nil` when there's no *current* auth_group ER → crash. A `nil` guard returning 404 would make this future-proof.

lib/dbservice/users.ex
Bahugunajii: Fixed — `get_student_by_student_id/2` no longer uses `Repo.get_by` (c66d8b4); it does `order_by: [asc: id], limit: 1` so it returns the earliest match instead of raising `MultipleResultsError`, and takes an optional auth_group to disambiguate. `get_student_by_id_or_apaar_id` routes its student_id branch through that same safe lookup. The remaining `Repo.one`-on-student_id (`duplicate_exists?`) is fixed in 0c82db4 via `Repo.exists?`.

lib/dbservice/users.ex
Bahugunajii: Fixed in 0c82db4 (Option A — in-code reactivation). On the create/update path the scoped lookup now passes `include_inactive: true`, so it matches the existing `(student_id, auth_group)` student even when a dropout flipped its ownership record to `is_current=false`. We then update that row and revive the record (`reactivate_auth_group_enrollment`) instead of inserting a duplicate. Added a test that drops a student then re-POSTs → same row, ownership current again.

Left the partial unique index as a possible fast-follow: ownership lives in `enrollment_record` (not a `student` column) so it can't be a plain unique index, and it'd need existing prod duplicates de-duped first.

lib/dbservice/users.ex
Bahugunajii: Fixed in 0c82db4 — `duplicate_exists?` now uses `Repo.exists?` instead of `Repo.one`, so a `student_id` shared by 2+ rows on the global fallback path returns a clean boolean instead of raising.

lib/dbservice/users.ex
Bahugunajii: Addressed in 0c82db4 (+ c66d8b4). `/enrolled`, `student_update_service.update_student_by_student_id`, and `student_profile_controller.create` now pass `auth_group` (from params) through to `get_student_by_student_id`, so they scope when an auth group is present and otherwise fall back to the now-non-raising, earliest-match unscoped lookup.

`user_session_controller.cleanup_student` / `remove_batch_mapping` only receive `student_id` (no auth_group in the payload), so they stay unscoped — being explicit per your note. They no longer raise (deterministic earliest match); we can scope them if those payloads start carrying auth_group.

lib/dbservice/users.ex
Bahugunajii: Fixed in 0c82db4 — the scoped update path now runs `validate_incoming_apaar_matches` before updating, reusing the same `validate_apaar_id_match` the global path uses. So re-POSTing `(S1, G, apaar=A_new)` over an existing `apaar=A_old` is rejected with the "doesn't match existing" error instead of silently overwriting. Added a test.

lib/dbservice_web/controllers/student_controller.ex
Bahugunajii: Fixed in 0c82db4 — `enrolled/2` now matches the lookup result and returns 404 ("Student not found with the provided identifier") when it's `nil`, instead of crashing on `student.user_id`. Body moved into a private `do_enrolled/3`.

Bahugunajii:
Thanks for the careful pass @suryabulusu — all six addressed (replied inline):

- **:466** duplicate-on-re-enroll → in-code reactivation: scoped create/update now matches the existing `(student_id, auth_group)` student even when a dropout deactivated its ownership record, and revives it instead of inserting a duplicate.
- **:503** `duplicate_exists?` → `Repo.exists?` (no more `MultipleResultsError` on the global path).
- **:475** scoped update now re-runs the apaar-matches-existing check before writing.
- **:277** `/enrolled` returns 404 on a `nil` lookup instead of crashing.
- **:197** `student_update_service` and `student_profile` now pass `auth_group` through; the two `user_session` admin endpoints stay unscoped (no auth_group in their payload) but no longer raise.
- **:180** already handled by the duplicate-safe `get_student_by_student_id`.

In `0c82db4` (the `get_student_by_student_id` safety landed in `c66d8b4`), with tests for the reactivation and apaar-overwrite guard. The partial unique index is noted as a possible fast-follow.

suryabulusu (approved):

Issue #546, opened 2026-06-12
Multiple repetitions are found for some chapter & topic codes in the database


ankurcdoshi:
I would suggest applying `unique` constraint on code fields in both `chapter` & `topic` tables to prevent this issue

PR #562, opened 2026-06-22
fix(chapter,topic): reject duplicate code values at the changeset level

Bahugunajii:
Refs #546 (Phase 1)

## Why

`code` values are duplicated for some `chapter` and `topic` rows in production. The end goal (per #546) is a DB-level unique index on `code`, but that **can't be added yet** — a `create unique_index` migration would crash on deploy the moment it hits an existing duplicate, and merging those duplicates is manual work (checking each row's associations first).

So this is **Phase 1**: stop *new* duplicates immediately, without touching existing data and without any migration that could break the deploy.

## What

Adds an application-level uniqueness check on `code` to both the `Chapter` and `Topic` changesets via `unsafe_validate_unique`. Because both the `resource`/chapter/topic APIs **and** the CSV import worker go through these changesets, this covers all the normal write paths.

The check is guarded to only run when a `code` is actually supplied (`get_change` returns a non-empty value), since the column is still nullable for now — rows without a code are left alone, and `unsafe_validate_unique` already excludes the current row on updates.

### Known limitations (by design)

This is a **stopgap, not the final guarantee**:
- Not race-safe — there's a small TOCTOU window between the check and the insert (acceptable here: chapters/topics are admin/import-created, low concurrency).
- Doesn't cover raw SQL / direct DB writes.

The **DB-level unique index** (plus making `code` `NOT NULL`) remains the eventual guarantee. That's **Phase 2**, which depends on first manually merging the existing duplicates and backfilling any null codes — to be tracked separately. Once the index exists, this validation becomes fully race-safe (and we'll pair it with `unique_constraint/2`).

## Tests

New `test/dbservice/chapters_test.exs` and `test/dbservice/topics_test.exs` cover, for each table:
- duplicate `code` on create → rejected with `"has already been taken"`
- multiple rows **without** a code → allowed
- updating a row without changing its code → allowed (no self-collision)
- updating a row to a code used by another row → rejected

`mix test` → 8/8. `mix format` + `mix credo` clean.

### DIFF
--- a/test/dbservice/chapters_test.exs
+++ b/test/dbservice/chapters_test.exs
@@ -3,6 +3,7 @@ defmodule Dbservice.ChaptersTest do
 
   alias Dbservice.Chapters
   alias Dbservice.Chapters.Chapter
+  alias Dbservice.Repo
 
   describe "chapter code uniqueness" do
     test "rejects creating a chapter with a code that already exists" do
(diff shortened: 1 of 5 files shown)


test/dbservice/chapters_test.exs
```diff
@@ -0,0 +1,35 @@
...
+    test "rejects creating a chapter with a code that already exists" do
+      assert {:ok, %Chapter{}} = Chapters.create_chapter(%{"code" => "CH-DUP"})
+
+      assert {:error, changeset} = Chapters.create_chapter(%{"code" => "CH-DUP"})
+      assert "has already been taken" in errors_on(changeset).code
+    end
+
+    test "allows multiple chapters without a code" do
+      assert {:ok, %Chapter{}} = Chapters.create_chapter(%{"name" => [%{"chapter" => "A"}]})
+      assert {:ok, %Chapter{}} = Chapters.create_chapter(%{"name" => [%{"chapter" => "B"}]})
```
ankurcdoshi: Can we stop this as well - creating any chapter/topic without a code?

ankurcdoshi (changes requested):

ankurcdoshi (approved):

PR #563, opened 2026-06-24
Add lms_teacher_feedback table (Teacher Feedback)

suryabulusu:
## What

Canonical schema for the **Teacher Feedback** feature: one row per teacher per setup. A Program Manager sets up a student-feedback round for a centre + batch and a set of teachers; the LMS records the mapping here and creates one db-service `session` per teacher.

Migration-only, following the `lms_pm_school_visits` pattern — this is operational, **LMS-owned** data that the LMS app reads/writes directly via the `pg` pool. There is no Ecto schema/context/controller in db-service for it. (Same ownership split as Holistic Mentorship: db-service owns the DDL, af_lms owns the DML — ADR-0005.)

## Schema

`lms_teacher_feedback`:

- **Identity/grouping**: `setup_run_id` (uuid; one per "Set Up" click), `cycle_label` ("Jun 2026").
- **Scope**: `school_code`, `centre_id`, `program_id`, `batch_class_ids` (text[]).
- **Teacher**: `teacher_id` (nullable for the free-text fallback), `teacher_name`, `teacher_order`.
- **Link to the session**: `session_pk`. The quiz id (= `session.platform_id`) and portal/admin links are filled on the session by the sessionCreator Lambda and resolved by joining `session_pk` at read time — not duplicated here.
- **Lifecycle**: `status` ∈ `pending`/`created`/`failed` (each teacher's setup can fail independently), `start_time`/`end_time` (UTC — see the note in the migration, `session` stores the same window in IST), `created_by`, soft-delete `deleted_at`, timestamps.

Indexes on `school_code`, `centre_id`, `program_id`, `setup_run_id`, `session_pk`, `(school_code, cycle_label)`, a partial active index, and a **partial unique index** on `(setup_run_id, teacher_order)`.

## Why the scope columns are what they are

A round's cohort is defined by the centre's **programme**, not by the school. A school can host both a CoE and a Nodal centre, each running its own batches, and the LMS selects a round's batches by `(school, centre programme)`. So both `centre_id` and `program_id` are recorded, and a historical row stays meaningful even if a centre is later renamed or re-pointed between programmes.

Display names are **not** stored — `centre_name` is resolved by joining `centres` on `centre_id` at read time, so a renamed centre reads correctly on old rounds instead of keeping a stale copy.

No FKs on `centre_id` / `program_id`: these are operational LMS rows written directly by the app, deliberately decoupled from db-service's referential graph, matching `lms_pm_school_visits`. Their **types** match their referents (both `bigint`) so joins need no casts.

## Update: five changes since first review

All made before this migration ran anywhere — staging's `schema_migrations` confirms it was never applied, so none of this needs a follow-up migration.

| Change | Why |
|---|---|
| **+ `program_id` (bigint)** | The round's cohort discriminator. Without it the cohort had to be re-inferred from the centre's *current* `program_id`, so re-pointing a centre silently changed what a past round meant. |
| **`centre_id`, `session_pk` `:integer` → `:bigint`** | They reference `centres.id` / `session.id`, both bigint. Wrong type against the referent, and forced a cast on every join. Also the class of bug the LMS scaffold records costing 3 days of prod 403s on document upload. |
| **+ unique index `(setup_run_id, teacher_order) WHERE deleted_at IS NULL`** | Setup writes a row per teacher in a loop with no uniqueness guarantee, so a double-submit created a second full set of rows *and* a second set of sessions — every teacher listed twice and the report's per-parameter denominator doubled. `teacher_order` is the per-setup key since `teacher_id` is nullable. Partial, so a soft-deleted row frees its slot. |
| **− `centre_name`** | Denormalised display name with no FK; a renamed centre left every historical row stale. Now joined from `centres`, which also stops `centre_id` being write-only. |
| **− `source_id`** | Never read back. Derivable from `(school_code, cycle_label)` and already on the session's `meta_data.cms_test_id`, which is where the BigQuery rows get it. |

The original description also listed `grade` and `batch_parent_id`; both were dropped earlier on this branch (`db14a7f`) and were never in the migration as it stands.

## Verified locally

- Migration applies cleanly and produces the intended schema (`centre_id`, `program_id`, `session_pk` all `bigint`).
- The app's real success-row and failed-row `INSERT`s both succeed.
- The unique index **rejects** a duplicate `(setup_run_id, teacher_order)` and **permits** reuse after a soft delete.
- The `status` check constraint still rejects bad values.

## Notes

- Pairs with af_lms #135 (which writes `program_id` and joins `centres` for the name) and the sessionCreator change (etl-data-flow #109).
- **Apply to staging and prod before the LMS feature is exercised there** — the LMS writes this table directly.

### DIFF
--- a/priv/repo/migrations/20260622120000_create_lms_teacher_feedback_table.exs
+++ b/priv/repo/migrations/20260622120000_create_lms_teacher_feedback_table.exs
@@ -46,7 +46,11 @@ defmodule Dbservice.Repo.Migrations.CreateLmsTeacherFeedbackTable do
       # Lifecycle: each teacher's setup can succeed or fail independently
       add :status, :string, size: 20, default: "pending", null: false
 
-      # Response window (stored UTC)
+      # Response window, stored in UTC (this table's convention, like the
+      # timestamps below). NOTE: the db-service `session` table stores the same
+      # window in IST (its legacy convention), so these will read ~5.5h earlier
+      # than session.start_time/end_time for the same round — do NOT compare them
+      # raw. The LMS UI tags these as UTC and renders in the viewer's timezone.
       add :start_time, :naive_datetime
       add :end_time, :naive_datetime
 
--- a/priv/repo/migrations/20260622120000_create_lms_teacher_feedback_table.exs
+++ b/priv/repo/migrations/20260622120000_create_lms_teacher_feedback_table.exs
@@ -27,10 +27,7 @@ defmodule Dbservice.Repo.Migrations.CreateLmsTeacherFeedbackTable do
       add :school_code, :string, size: 20, null: false
       add :centre_id, :integer
       add :centre_name, :string, size: 255
-      add :batch_parent_id, :string, size: 255
       add :batch_class_ids, {:array, :string}, default: [], null: false
-      # grade is informational only; a feedback round can span grades.
-      add :grade, :integer
 
       # Teacher (id nullable: free-text fallback when no roster id is available)
       add :teacher_id, :string, size: 50
--- a/priv/repo/migrations/20260622120000_create_lms_teacher_feedback_table.exs
+++ b/priv/repo/migrations/20260622120000_create_lms_teacher_feedback_table.exs
@@ -35,10 +35,11 @@ defmodule Dbservice.Repo.Migrations.CreateLmsTeacherFeedbackTable do
       add :teacher_name, :string, size: 255, null: false
       add :teacher_order, :integer, null: false
 
-      # Created artifacts (quiz-backend quiz + db-service session)
-      add :quiz_id, :string, size: 255
+      # The db-service session this teacher's feedback round created. The
+      # sessionCreator Lambda fills the quiz_id (session.platform_id) + portal /
+      # admin links onto the SESSION row asynchronously; we resolve them by
+      # joining on session_pk at read time, so they are not duplicated here.
       add :session_pk, :integer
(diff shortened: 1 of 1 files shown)


priv/repo/migrations/20260622120000_create_lms_teacher_feedback_table.exs
```diff
@@ -0,0 +1,65 @@
...
+  # Operational LMS-owned data like lms_pm_school_visits — the LMS app reads and
+  # writes it directly, so there is no schema/context/controller here.
+  def change do
+    create table(:lms_teacher_feedback) do
+      add :setup_run_id, :uuid, null: false
+      add :cycle_label, :string, size: 50, null: false
+
+      # A round's cohort is (school, centre programme) — a school can host both a
+      # CoE and a Nodal centre. bigint matches the referents; no FKs (see above).
+      add :school_code, :string, size: 20, null: false
```
Bahugunajii, line 14: Shouldn't we just take reference from school table?

priv/repo/migrations/20260622120000_create_lms_teacher_feedback_table.exs
```diff
@@ -0,0 +1,65 @@
...
+      add :centre_id, :bigint
+      add :program_id, :bigint
```
Bahugunajii: Same, take reference from "centre" & "program"

Bahugunajii (approved):

PR #564, opened 2026-06-25
Range filters on GET session 

Abhinav-3009:
Add `end_time` range filters to `GET /api/session`

  Support optional `end_time_gte` (inclusive) and `end_time_lt` (exclusive) query params on the session index endpoint, callers can fetch sessions within an end-time window. Used by the quiz ETL incremental-sync flow to discover recently-finalized quiz sessions.

  - Add both params to the index Swagger spec.
  - Refactor index query building into build_session_index_query/1 using reduce_while so a bad timestamp can short-circuit with an error.
  - Parse range values with DateTime.from_iso8601; invalid timestamps return HTTP 400 {"error": "Invalid end_time range timestamp"}.
  - Filters are additive: existing equality, pagination, and metadata filters are unchanged, so callers that don't pass the new params behave as before.
  - Add controller tests: platform equality still works, platform + end_time range (inclusive lower / exclusive upper bounds, non-quiz excluded), and invalid timestamp returns 400.

### DIFF
--- a/lib/dbservice_web/controllers/session_controller.ex
+++ b/lib/dbservice_web/controllers/session_controller.ex
@@ -5,6 +5,7 @@ defmodule DbserviceWeb.SessionController do
   alias Dbservice.Repo
   alias Dbservice.Sessions
   alias Dbservice.Sessions.Session
+  alias Dbservice.Utils.Util
 
   action_fallback DbserviceWeb.FallbackController
 
@@ -110,7 +111,7 @@ defmodule DbserviceWeb.SessionController do
   end
 
   defp apply_end_time_filter(operator, value, acc) do
-    case parse_datetime(value) do
+    case Util.parse_datetime(value) do
       {:ok, datetime} ->
         query =
           case operator do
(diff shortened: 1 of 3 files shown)


lib/dbservice_web/controllers/session_controller.ex
```diff
@@ -83,6 +109,31 @@ defmodule DbserviceWeb.SessionController do
...
+
+  defp parse_datetime(value) when is_binary(value) do
+    case DateTime.from_iso8601(value) do
+      {:ok, datetime, _offset} -> {:ok, DateTime.truncate(datetime, :second)}
+      _ -> :error
+    end
+  end
+
+  defp parse_datetime(_value), do: :error
+
```
Bahugunajii: I think we have a utility file maintained for these functions. Maybe reuse them (if exists) otherwise maybe shift these helper functions to that file

Bahugunajii (changes requested):

lib/dbservice_web/controllers/session_controller.ex
Abhinav-3009: Moved parse_datetime into Dbservice.Utils.Util, and the controller now calls Util.parse_datetime

Bahugunajii (approved):

PR #568, opened 2026-06-30
fix(resources): return test problems in type_params order

### LINKED ISSUE #565: Test problems are not coming in same order as defined in test
Test problems' order is defined in `resource` table --> `type_params` column. Test problems api should return problems in that same order. Right now api is returning problems in ascending order of their ids.

Example Test problems api -
```
http://localhost:4000/api/resource/test/2228/problems?lang_code=en&curriculum_id=1
```

`type_params` value for this test in db is as follows.
```
{
  "marks": 180,
  "duration": "45",
  "subjects": [
    {
      "marks": 180,
      "sections": [
        {
          "name": "MCQ Single Answer",
          "type": "mcq_single_answer",
          "marks": 180,
          "compulsory": {
            "problems": [
              {
                "id": 2038
              },
              {
                "id": 2040
              },
              {
                "id": 2220
              },
              {
                "id": 2044
              }
            ]
          }
        }
      ],
      "subject_id": 2
    }
  ]}
```
But right now api is returning these problems in order - 2038, 2040, 2044, 2220

Bahugunajii:
Fixes #565

## Why

`GET /api/resource/test/:id/problems` returned problems in **ascending id order** instead of the order defined in the test's `resource.type_params`.

`extract_problem_ids_from_test/1` already produces the ids in `type_params` order, but `fetch_and_format_problems` then ran `where r.id in ^problem_ids` + `Repo.all` — and SQL `IN` does **not** preserve the list order, so the rows came back effectively in id order. From the issue: type_params order `2038, 2040, 2220, 2044` was returned as `2038, 2040, 2044, 2220`.

## What

Reorder the fetched problem resources to match the `type_params` id order before building the response (`order_problems_by_ids/2`):
- ids that don't resolve to a `problem` resource are dropped,
- duplicate ids collapse to their first occurrence.

No API/contract change — only the ordering of the returned list is corrected.

## Verification

- Added a test in `resources_test.exs`: builds a test whose `type_params` lists 4 problems in a deliberately scrambled order (vs their ascending ids) and asserts the response comes back in `type_params` order. Fails on `main`, passes with this fix.
- `mix test test/dbservice/resources_test.exs` → 6/6; `mix credo` clean; format clean.

### DIFF
--- a/lib/dbservice/resources.ex
+++ b/lib/dbservice/resources.ex
@@ -93,6 +93,9 @@ defmodule Dbservice.Resources do
         ]
       )
       |> Repo.all()
+      # SQL `IN` does not preserve list order, so re-sort the fetched problems to match the
+      # order the ids appear in the test's type_params (the order the test defines).
+      |> order_problems_by_ids(problem_ids)
 
     # topic_id lives on the resource_topic join table, not on the resource
     # itself, so fetch the mapping separately and attach it per problem.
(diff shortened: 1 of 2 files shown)


ankurcdoshi (approved):
Tested to be working.

Draft PR [https://github.com/avantifellows/db-service/pull/566](https://github.com/avantifellows/db-service/pull/566) by Deepansh also seems to be fixing same bug.

PR #569, opened 2026-06-30
perf(db): add user_id and session_id indexes on user_session

### LINKED ISSUE #493: Add missing user_id and session_id indexes on user_session (largest table)
## What's happening now

The `user_session` table is the **largest table in the database** (3.9M rows, 407 MB) and is missing indexes on two of its most-queried columns: `user_id` and `session_id`. Every lookup by either column triggers a **parallel sequential scan** across the entire table.

**EXPLAIN ANALYZE from production (user_id lookup):**

```
Parallel Seq Scan on user_session
  Filter: (user_id = 297513)
  Rows Removed by Filter: 1,301,272 (per worker, 3 workers)
Execution Time: 243.702 ms
```

Three parallel workers each scan ~1.3M rows, reading the entire 407 MB table, just to find one user's sessions.

**Audit reference:** `db-investigation.md`, Section 2.4 (F13), Tier 1 Action #3. Also relates to F18 (missing FK indexes on `session_id`).

**Confirmed by:** Codex (GPT-5.5), Gemini, Kimi — 2/3 rate Critical, 1 rates High

---

### The index was accidentally dropped during a migration

Kimi traced the exact history through the migration files:

1. **`20220429194405`** — Created `user_session` table with `user_id` column and **explicitly created an index on it**
2. **`20230320064301`** — Removed `user_id` from the table, which **implicitly dropped the index**
3. **`20240525065205`** — Re-added `user_id` using `add :user_id, references(:user)`

The problem is in step 3: Ecto's `references()` creates a **foreign key constraint** but does **not** recreate the index. The index was silently lost in production over a year ago.

### Multiple code paths hit this unindexed column

| Location | Query Pattern | Impact |
|---|---|---|
| `user_session_controller.ex:197` | `Repo.exists?(from u in UserSession, where: u.user_id == ^user_id)` | Called during student cleanup — 244ms per check |
| `sessions.ex:335` | `from(us in UserSession, where: us.user_id == ^user_id) \|> Repo.delete_all()` | Called during teacher/student deletion — full table scan to find rows to delete |
| `user_session_controller.ex:44-58` | Dynamic `index` endpoint filtering by `user_id` | API list endpoint — 244ms per request |
| `user.ex:36` | `many_to_many(:sessions, SessionOccurrence, join_through: "user_session")` | Association preloads — every `Repo.preload(:sessions)` on a User triggers this scan |

### `session_id` is also unindexed

The same table is also missing an index on `session_id`. This column is used for the reverse direction of the many-to-many relationship (fetching users for a given session) and for FK constraint checks. This is tracked as F18 in the audit (37 FK columns lacking indexes).

---

### Create a single migration that adds both indexes

```elixir
defmodule Dbservice.Repo.Migrations.AddUserSessionIndexes do
  use Ecto.Migration

  @disable_ddl_transaction true
  @disable_migration_lock true

  def change do
    create_if_not_exists index(:user_session, [:user_id], concurrently: true)
    create_if_not_exists index(:user_session, [:session_id], concurrently: true)
  end
end
```

Breaking this down:
- `@disable_ddl_transaction true` — required for `CREATE INDEX CONCURRENTLY`. Without this, Ecto wraps the migration in a transaction, and PostgreSQL does not allow concurrent index builds inside transactions.
- `@disable_migration_lock true` — prevents Ecto from acquiring an advisory lock that could conflict with the concurrent build.
- `concurrently: true` — generates `CREATE INDEX CONCURRENTLY`, which builds the index **without locking the table for writes**. Reads and inserts continue normally during the build.
- `create_if_not_exists` — idempotent, safe to re-run if the migration is interrupted.

### Why a simple single-column index is sufficient

- The critical hot-path queries (`delete_user_sessions_by_user_id`, `check_if_session_accessed`, association preloads) filter **only** by `user_id`
- The controller's dynamic query builder filters by single fields individually, not by compound predicates
- `session_occurrence_id` already has its own index from the original migration
- A composite index would be larger and wouldn't help the `delete_all` or `exists?` queries that only use `user_id`

### Expected index sizes and write overhead

| Index | Estimated Size | Rationale |
|---|---|---|
| `user_id` | 60–120 MB | B-tree on bigint, 3.9M rows. PostgreSQL 14 B-tree deduplication helps since many users have multiple sessions |
| `session_id` | 60–120 MB | Similar structure |

**Write overhead is minimal.** The audit confirms `user_session` is effectively insert-only: 3,835,707 inserts, 0 updates, 49 deletes since stats were reset. Each new insert will maintain two additional B-tree entries — trivial compared to the current 244ms read penalty.

### Deployment checklist

1. **Increase `maintenance_work_mem`** before running the migration (if #488 hasn't been applied yet): `SET maintenance_work_mem = '512MB';` — allows the index build to sort 3.9M tuples in memory instead of spilling to disk
2. **Schedule during low-traffic hours** — the concurrent build does two full table scans and adds CPU/I/O load. On a `db.t4g.large`, expect 2–5 minutes per index
3. **Monitor during build:**
   - `SELECT * FROM pg_stat_progress_create_index;` — track progress
   - RDS `ReadIOPS` and `CPUUtilization` in CloudWatch
   - WAL/replica lag if replicas exist
4. **Verify indexes are valid after build:**
   ```sql
   SELECT indexrelid::regclass, indisvalid
   FROM pg_index
   WHERE indrelid = 'user_session'::regclass;
   ```
   `CONCURRENTLY` can leave an index in an invalid state if it fails (e.g., due to a deadlock or unique violation). If `indisvalid = false`, drop and recreate.
5. **Run ANALYZE:** `ANALYZE user_session;` — refresh statistics so the planner uses the new indexes
6. **Re-run EXPLAIN ANALYZE** on the F13 query to confirm the parallel seq scan is gone

---

## Severity

**Critical** — confirmed by 2/3 council reviewers. This is a 244ms full-table scan on the largest table in the database, caused by an accidentally dropped index. The fix is a straightforward index creation with zero application code changes.

Bahugunajii:
Closes #493

## Why

`user_session` (~3.9M rows, 407 MB — the largest table) has no index on `user_id` or `session_id`, so every lookup is a parallel sequential scan (~244ms in prod). The `user_id` index was implicitly dropped when the column was removed (`20230320064301`) and re-added via `references/2` (`20240525065205`), which creates the FK constraint but **not** an index.

Hot paths hitting these unindexed columns: `Sessions.delete_user_sessions_by_user_id` (`delete_all`), `user_session_controller` cleanup `exists?`, the dynamic list endpoint, and `User` ↔ `SessionOccurrence` association preloads.

## What

New migration adding single-column indexes on `user_id` and `session_id`, built **`CONCURRENTLY`** (`@disable_ddl_transaction true` + `@disable_migration_lock true`) so the table isn't write-locked during the build. `create_if_not_exists` keeps it idempotent if interrupted.

App code unchanged.

## Deploy note
Concurrent build does two full scans of a 407 MB table — run during low traffic (~2–5 min/index on `db.t4g.large`), then `ANALYZE user_session;`. If a build is interrupted, `CONCURRENTLY` can leave an invalid index — check `pg_index.indisvalid` and drop/recreate if needed.

### DIFF
--- a/priv/repo/migrations/20260701120000_add_user_session_indexes.exs
+++ b/priv/repo/migrations/20260701120000_add_user_session_indexes.exs
@@ -1,0 +1,17 @@
+defmodule Dbservice.Repo.Migrations.AddUserSessionIndexes do
+  use Ecto.Migration
+
+  # CREATE INDEX CONCURRENTLY cannot run inside a transaction, and the migration
+  # advisory lock can conflict with the concurrent build, so both are disabled.
+  @disable_ddl_transaction true
+  @disable_migration_lock true
+
+  # `user_session` is the largest table (~3.9M rows) and was left without indexes on
+  # `user_id`/`session_id` after `user_id` was dropped and re-added via references/2
+  # (which creates the FK but not an index). Both columns are queried on hot paths
+  # (deletes, exists? checks, association preloads), causing full table scans.
+  def change do
+    create_if_not_exists index(:user_session, [:user_id], concurrently: true)
+    create_if_not_exists index(:user_session, [:session_id], concurrently: true)
+  end
+end


suryabulusu (approved):

deepansh96:
Could we split these into two migrations, with one index in each, and use `create` instead of `create_if_not_exists`?

A concurrent index build can leave an invalid index behind if it is interrupted. On retry, `create_if_not_exists` only checks the index name, so it may skip that invalid index and allow the migration to appear successful.

Ecto already skips migrations that completed successfully. With separate migrations, if the second index fails, only that migration needs to be retried. Using `create` also makes the broken index visible, so it can be dropped and rebuilt instead of being silently skipped.

We should also check `pg_index.indisvalid` after deployment.

PostgreSQL reference: https://www.postgresql.org/docs/15/sql-createindex.html

deepansh96:
we should not disable the migration lock. if we split this into two migrations, even then, we should remove `@disable_migration_lock true` and keep the ddl disable one. we should also configure the repo to use `migration_lock: :pg_advisory_lock`. this is what's recommended by ecto docs https://ecto-sql.hexdocs.pm/3.12.0/Ecto.Migration.html#index/3-adding-dropping-indexes-concurrently . because this is something we can control and have to do only once, we should do it during low traffic and monitor it while it runs. with `concurrently: true`, we should not have to stop the existing traffic.

image

PR #570, opened 2026-06-30
perf(db): add composite index on session_occurrence(session_id, start_time, end_time)

### LINKED ISSUE #492: Add composite index on session_occurrence (session_id, start_time, end_time)
## What's happening now

The `session_occurrence` query filtered by `session_id` and a time range is the **#3 cumulative time consumer** in the database: 2.05 million calls, 39,748 seconds (11.0 hours) total, 19.35ms historical mean.

The query:

```sql
SELECT ... FROM session_occurrence
WHERE start_time <= $1 AND end_time >= $2 AND session_id = $3
ORDER BY id LIMIT/OFFSET
```

The table has separate single-column indexes on `session_id`, `start_time`, and `end_time`, but **no composite index** covering all three columns. PostgreSQL combines two indexes using a BitmapAnd operation — functional but suboptimal.

**Audit reference:** `db-investigation.md`, Section 2.3b (F12), Tier 1 Action #4

**Confirmed by:** Codex (GPT-5.5), Gemini — both recommend this as a separate issue from #491 (ORDER BY fix)

---

### BitmapAnd is a workaround, not an optimal plan

The current EXPLAIN ANALYZE shows:

```
BitmapAnd
  Bitmap Index Scan on session_occurrence_session_id_index (42 rows)
  Bitmap Index Scan on session_occurrence_start_time_index (4,750 rows)
Bitmap Heap Scan (Rows Removed by Filter: 2)
Execution Time: 0.278 ms
```

Breaking this down:
- PostgreSQL scans the `session_id` index and gets 42 row pointers
- It separately scans the `start_time` index and gets 4,750 row pointers
- It intersects these two bitmap sets (BitmapAnd) to find rows matching both conditions
- It then fetches those rows from the heap and applies the remaining `end_time` filter

The current sampled execution is fast (0.278ms), but the historical mean is 19.35ms across 2.05M calls. The BitmapAnd approach is sensitive to data distribution — when a session has many occurrences or the time range is wide, the bitmap intersection grows expensive. The 11 hours of cumulative DB time confirms this happens frequently under real workload conditions.

### A composite index eliminates the intersection step

With a composite index on `(session_id, start_time, end_time)`, PostgreSQL can satisfy the entire WHERE clause in a single B-tree traversal: jump to the session_id, then range-scan within the time window. No bitmap construction, no intersection, no heap re-check.

---

### Create the composite index

```sql
CREATE INDEX CONCURRENTLY ON session_occurrence (session_id, start_time, end_time);
```

Breaking this down:
- `session_id` first — the equality predicate. Narrows to all occurrences for one session (typically tens to low hundreds of rows).
- `start_time` second — enables a range scan within the session's occurrences.
- `end_time` third — included in the index so the planner can apply the `end_time >= X` condition as an index filter without going to the heap.
- `CONCURRENTLY` — builds without locking the table for writes.

### Consider dropping the duplicate session_id index

The table has a duplicate index `session_occurence_session_id_index` (note the typo — "occurence" vs "occurrence"). The audit identifies this as finding F19:

> Duplicate `session_occurrence` index (typo, 11 MB)

Once the composite index is in place, the standalone `session_id` index is fully redundant — the composite's leading column serves all `WHERE session_id = ?` queries. Monitor `pg_stat_user_indexes` for a week after creating the composite, then drop the standalone:

```sql
DROP INDEX CONCURRENTLY session_occurence_session_id_index;
```

This reclaims 11 MB and eliminates write amplification from maintaining a redundant index.

### Application-level improvement (medium-term)

The controller uses a CASE-based query pattern that merges continuous and non-continuous schedule queries into one code path. These have different access patterns:

- **Continuous schedule:** `WHERE session_id = ? AND start_time <= ? AND end_time >= ?` — the temporal range query this index optimizes
- **Non-continuous schedule:** Different filter logic

Splitting these into two distinct Ecto queries allows the planner to generate optimal plans for each, rather than a compromise plan that handles both.

---

## Severity

**Critical** — #3 cumulative time consumer (11 hours total). The composite index is a low-risk, high-impact database migration that eliminates the BitmapAnd overhead and collapses three separate index lookups into one.

Bahugunajii:
Closes #492

## Why

The active-window query `WHERE session_id = ? AND start_time <= ? AND end_time >= ? ORDER BY id` is one of the top cumulative-time consumers (≈11h total over ~2M calls). The table has separate single-column indexes on `session_id`, `start_time`, `end_time`, so the planner does a **BitmapAnd** intersection — functional but sensitive to data distribution (slow when a session has many occurrences or the range is wide).

## What

Adds a composite index `(session_id, start_time, end_time)`:
- `session_id` (equality) narrows to one session,
- `start_time` enables the range scan,
- `end_time` lets the planner apply `end_time >= ?` from the index.

Built **`CONCURRENTLY`** (`@disable_ddl_transaction` + `@disable_migration_lock`), `create_if_not_exists` for idempotency. No app code change.

## Follow-ups noted in the issue (not in this PR)
- After ~1 week of `pg_stat_user_indexes` monitoring, the redundant standalone `session_occurence_session_id_index` (note the typo, ~11 MB) can be dropped.
- Splitting the continuous vs non-continuous schedule queries into distinct Ecto queries is a separate app-level improvement.

### DIFF
--- a/priv/repo/migrations/20260701120100_add_session_occurrence_composite_index.exs
+++ b/priv/repo/migrations/20260701120100_add_session_occurrence_composite_index.exs
@@ -1,0 +1,16 @@
+defmodule Dbservice.Repo.Migrations.AddSessionOccurrenceCompositeIndex do
+  use Ecto.Migration
+
+  @disable_ddl_transaction true
+  @disable_migration_lock true
+
+  # The active-window query `WHERE session_id = ? AND start_time <= ? AND end_time >= ?`
+  # currently relies on a BitmapAnd of separate single-column indexes (the #1/#3 cumulative
+  # time consumers). A composite (session_id, start_time, end_time) lets the planner satisfy
+  # the whole predicate in one B-tree traversal: equality on session_id, then a range scan.
+  def change do
+    create_if_not_exists index(:session_occurrence, [:session_id, :start_time, :end_time],
+                           concurrently: true
+                         )
+  end
+end


suryabulusu (approved):

deepansh96:
The same comments from #569 apply here as well: [`create_if_not_exists` with a concurrent index build](https://github.com/avantifellows/db-service/pull/569#issuecomment-5035533743), and [disabling the migration lock](https://github.com/avantifellows/db-service/pull/569#issuecomment-5035708371).

Bahugunajii:
Addressed both in 52944bd:

- **`create_if_not_exists` → `create`** — so an interrupted concurrent build leaves a *visible* invalid index (rather than being silently skipped by the name check) that can be dropped and rebuilt. This PR only adds one index, so no split was needed. Will check `pg_index.indisvalid` after deploy.
- **Migration lock** — removed `@disable_migration_lock true` (kept `@disable_ddl_transaction true`) and set `migration_lock: :pg_advisory_lock` on `Dbservice.Repo` in `config/config.exs`, per the Ecto concurrent-index docs. The migration lock is now held via an advisory lock that doesn't block the concurrent build.

Verified on a local DB: migration runs and the index comes out `indisvalid = true`. Plan is still to run this during low traffic and monitor.

deepansh96 (approved):

PR #571, opened 2026-06-30
perf(db): add composite index on group_user(user_id, group_id)

### LINKED ISSUE #489: Add composite index on group_user and reduce 4.58M membership probe volume
## What's happening now

The `group_user` table is the **#1 cumulative time consumer** in the entire database: 4.58 million calls, 82,938 seconds (23.0 hours) of total database time, with an 18.12ms historical mean and a 508ms worst case.

The query pattern is a membership check:

```sql
SELECT ... FROM group_user WHERE group_id = $1 AND user_id = $2 ORDER BY id LIMIT/OFFSET
```

The table has two separate single-column indexes (`group_user_group_id_index` and `group_user_user_id_index`) but **no composite index on both columns**. PostgreSQL picks one index and filters on the other column, which means it scans rows it doesn't need.

**Audit reference:** `db-investigation.md`, Section 2.1 (F9), Tier 1 Action #2

**Confirmed by:** Codex (GPT-5.5), Gemini, DeepSeek — 2/3 rate Critical, 1 rates High

---

### The per-query cost is variable and can spike badly

The current sampled EXPLAIN ANALYZE shows 0.083ms execution, which looks fine. But `pg_stat_statements` reports an 18.12ms mean across 4.58M calls and a 508ms maximum. The difference is explained by **group size skew**:

| group_id | member count |
|----------|-------------|
| 8851 | 258,977 |
| 8850 | 155,725 |
| 8085 | 134,738 |
| 3 | 117,793 |
| 2 | 71,465 |

When PostgreSQL uses the `user_id` index to probe a user in one of these giant groups, the filter step (`Filter: group_id = $1`) scans many rows. When it happens to choose the `group_id` index for a 258K-member group, it scans hundreds of thousands of rows to find one user. The 18ms historical mean reflects this reality across 10 months of production data.

### The call volume itself is a problem

4.58 million membership probes means the application is checking "does this user belong to this group?" individually, repeatedly, across multiple code paths. Even with a perfect index, millions of unnecessary DB round-trips waste connection pool time, network latency, and CPU.

Three specific code paths drive this volume:

**1. `batch_enrollment_service.ex` — per-row DB queries in `Enum.find`**

`group_user_by_type?/2` fires a separate `Repo.exists?` query against the `group` table for every `group_user` row, called inside `Enum.find`. A user with 12 group memberships triggers up to 12 individual DB queries just to find their batch group.

**2. `enrollment_service.ex` — 4 individual lookups per student enrollment**

Every enrollment calls `get_group_user_by_user_id_and_group_id` separately for each group type (batch, grade, school, auth_group). Enrolling N students generates 4N individual lookups.

**3. `user_controller.ex` — session fetches probe group_user for batch access**

With 297M cumulative session lookups (F15), many trigger a `group_user` join to filter by batch membership. This is a major contributor to the 4.58M call volume.

---

### Phase 1 — Composite index (immediate, minutes to apply)

Create a composite index with `user_id` as the leading column. All three council reviewers recommend this order over the audit's original `(group_id, user_id)` because `user_id` is far more selective — users have 9-12 memberships, while groups have up to 258K members.

```sql
CREATE INDEX CONCURRENTLY ON group_user (user_id, group_id);
```

Breaking this down:
- `CONCURRENTLY` — builds the index without locking the table for writes. Takes longer but causes zero downtime.
- `user_id` first — PostgreSQL traverses fewer B-tree pages because the user_id prefix narrows to 9-12 rows immediately, vs group_id narrowing to up to 258K rows.
- This index also serves queries that filter on `user_id` alone (like `get_group_user_by_user_id`), making it dual-purpose.

**Consider making it UNIQUE** if duplicate `(user_id, group_id)` pairs are invalid in the application. A unique index would:
- Enforce data integrity at the database level
- Eliminate the need for `ORDER BY id LIMIT 1` in the query (there can only be one result)
- Be slightly more efficient for lookups

```sql
-- Use this instead if duplicates should be impossible:
CREATE UNIQUE INDEX CONCURRENTLY ON group_user (user_id, group_id);
```

**Do NOT drop the existing single-column indexes yet.** The `group_id` index is still needed for queries that list all members of a group. The `user_id` index may become redundant (covered by the composite), but monitor `pg_stat_user_indexes` for a full representative period before removing it.

### Phase 2 — Application-level batching (short-term)

**Fix 1: Replace `group_user_by_type?` with preloaded groups**

Instead of querying the `group` table per row inside `Enum.find`, preload the `group` association in a single query:

```elixir
# Fetch group_users with their group type in one query
def get_group_user_by_user_id_with_groups(user_id) do
  from(gu in GroupUser,
    join: g in Group, on: g.id == gu.group_id,
    where: gu.user_id == ^user_id,
    preload: [group: g]
  )
  |> Repo.all()
end
```

Then `Enum.find` becomes a pure in-memory check — zero DB queries:

```elixir
batch_group_user = Enum.find(group_users, fn gu -> gu.group.type == "batch" end)
```

**Fix 2: Batch enrollment lookups with `WHERE group_id IN (...)`**

Instead of 4 individual `get_group_user_by_user_id_and_group_id` calls per student, fetch all at once:

```elixir
def get_group_users_by_user_id_and_group_ids(user_id, group_ids) when is_list(group_ids) do
  from(gu in GroupUser,
    where: gu.user_id == ^user_id and gu.group_id in ^group_ids
  )
  |> Repo.all()
  |> Map.new(&{&1.group_id, &1})
end
```

This reduces 4 queries to 1 per student enrollment.

### Phase 3 — Application-level caching (medium-term)

For high-traffic paths like session fetches that probe batch membership:

- Cache `{user_id, MapSet<group_ids>}` in an ETS table with a 60–300 second TTL
- Batch/group membership changes infrequently (only on enrollment/re-enrollment events), so cache invalidation is straightforward
- This could cut the 4.58M DB probes by 80–95%

---

## Post-change validation

1. After creating the composite index, run `EXPLAIN ANALYZE` on the hot query to confirm it uses the new index
2. Monitor `pg_stat_user_indexes` for the new composite vs the old single-column indexes — track scan counts over 1 week
3. Re-check `pg_stat_statements` after one week — the cumulative time for this query pattern should drop significantly
4. For application-level fixes, measure the total `group_user` query count before and after — target is reducing from 4.58M to under 500K

---

## Severity

**Critical** — confirmed by 2/3 council reviewers. This is the #1 cumulative time consumer in the database (23 hours total). The composite index is a low-risk, high-impact immediate fix; the application-level changes address the root cause of the excessive call volume.

Bahugunajii:
Closes #489

## Why

The membership probe `WHERE group_id = ? AND user_id = ?` is the **#1 cumulative time consumer** (~23h over 4.58M calls, 18ms mean, 508ms worst case). The table has only separate single-column indexes, so the planner scans rows on one column and filters the other — which explodes for large groups (some have 100k–258k members).

## What

Adds a composite index `(user_id, group_id)` — `user_id` leading because it's far more selective (users have ~9–12 memberships vs groups up to 258k). It answers the probe in a single B-tree traversal and is dual-purpose (also serves `user_id`-only lookups like `get_group_user_by_user_id`).

Built **`CONCURRENTLY`**, `create_if_not_exists`. The existing single-column indexes are left in place (the `group_id` one still serves "list all members of a group").

**Non-unique on purpose:** the issue suggests *considering* a UNIQUE index, but that requires de-duping existing `(user_id, group_id)` pairs first (same class of work as #546/#482), so it's out of scope for this low-risk index add.

The issue's Phase 2/3 app-level batching + caching (reducing the 4.58M call volume) are separate follow-ups.

### DIFF
--- a/priv/repo/migrations/20260701120200_add_group_user_composite_index.exs
+++ b/priv/repo/migrations/20260701120200_add_group_user_composite_index.exs
@@ -1,0 +1,18 @@
+defmodule Dbservice.Repo.Migrations.AddGroupUserCompositeIndex do
+  use Ecto.Migration
+
+  @disable_ddl_transaction true
+  @disable_migration_lock true
+
+  # The membership probe `WHERE group_id = ? AND user_id = ?` is the #1 cumulative time
+  # consumer. With only separate single-column indexes, the planner picks one and filters on
+  # the other - scanning many rows for large groups (some have 100k+ members). A composite
+  # with user_id leading (users have ~9-12 memberships, far more selective than group_id)
+  # answers the probe in one traversal and also serves user_id-only lookups.
+  #
+  # Non-unique on purpose: a UNIQUE index would need existing duplicate (user_id, group_id)
+  # pairs de-duped first, which is tracked separately.
+  def change do
+    create_if_not_exists index(:group_user, [:user_id, :group_id], concurrently: true)
+  end
+end


suryabulusu (approved):

Bahugunajii:
Applied the same review guidance as #570 (from #569): switched `create_if_not_exists` → `create` so an interrupted concurrent build leaves a visible INVALID index (not silently skipped), removed `@disable_migration_lock true` (kept `@disable_ddl_transaction true`), and set `migration_lock: :pg_advisory_lock` on the repo per the Ecto concurrent-index docs. Verified locally: migration runs and the index is `indisvalid = true`.

Note: the one-line `migration_lock` config is added in #570/#572/#589 too — it'll conflict trivially on merge (keep one copy).

deepansh96 (approved):

PR #586, opened 2026-07-07
chore(infra): address pre-cutover review feedback

Bahugunajii:
Addresses the review feedback ahead of the prod ECS cutover. Grouped by area.

## ✅ Implemented

**Deploy workflows (staging + prod)**
- **Image tag from checked-out ref** — `IMAGE_TAG` now derives from `git rev-parse HEAD` instead of `${{ github.sha }}`, so image tag / source commit / task def stay aligned even when a specific ref is dispatched.
- **VPC-scoped SG lookup** — `describe-security-groups` now filters on `vpc-id` too (via a `VPC_ID` env), since SG names are only unique within a VPC.
- **Prod release-only** — the prod workflow now fails fast unless the deploy ref is `release`. Staging keeps the free-form ref for validation.

**Terraform**
- **Deployment circuit breaker** (`ecs.tf`) — `deployment_circuit_breaker { enable = true, rollback = true }` so a revision that can't reach steady state auto-rolls back. Matters because CI registers task-def revisions and the service has `ignore_changes = [task_definition]`, so a later apply won't self-correct a stuck deploy.
- **ECS Exec IAM** (`iam.tf`) — added the `ssmmessages:*Channel` permissions the **task role** needs; `enable_execute_command` was on but ECS Exec couldn't actually work without these.
- **Narrowed CI deploy IAM** (`iam.tf`) — ECR push/pull scoped to `db-service-*` repos; scopable ECS actions (`UpdateService`, `DescribeServices`, `RunTask`, `StopTask`, `DescribeTasks`) scoped to `db-service-*` cluster/service/task/task-def ARNs. Actions that AWS doesn't allow resource-level scoping on (`RegisterTaskDefinition`, `DescribeTaskDefinition`, `ListTasks`, `ecr:GetAuthorizationToken`, `ec2/elb Describe*`) are kept on `*` in their own statements, commented as such.
- **Commit `.terraform.lock.hcl`** — removed it from `.gitignore` and checked it in so staging/prod applies pin identical provider versions.
- **`terraform.tfvars.example`** — documents the full input set with placeholders (secrets noted as `TF_VAR_*`-only).
- **Makefile wrappers** — `tf-init/plan/apply-staging|prod` always pair the backend state key with the matching tfvars (and `-reconfigure` on init), preventing the "prod values against staging state" footgun. Staging targets include `create_ci_user=true` since the shared CI user lives in staging state with `prevent_destroy`.

**Dockerfile**
- `EXPOSE 8080` to match the ECS container port.

## ℹ️ Ops note
- **`DB_SERVICE_AWS_*` secrets** are already set on the repo (added during staging setup), so the staging workflow authenticates today. Same secrets serve prod. ✅

## Validation
- `terraform validate` ✓, `terraform fmt` ✓, `actionlint` ✓ on both workflows.
- ⚠️ The **CI IAM narrowing** and **circuit breaker** only take effect on `terraform apply`; recommend applying to staging and running one deploy to confirm the pipeline still authenticates end-to-end before prod cutover.


deepansh96:
Added commit `3d4a7d27` to both ECS deploy workflows. After ECS reports the service as stable, the workflow now checks that the active task definition is the one created by this run. Without this check, ECS could roll back a failed deploy to the old healthy version and the workflow could still show green. It will now fail clearly when a rollback happens.

deepansh96:
Added commit `90bf0464` so every Terraform plan/apply Make target runs the matching init target first. Terraform remembers the last selected state file; without this, production values could accidentally be applied against staging state, or the reverse. The targets now always select the correct state before planning or applying.

PR #587, opened 2026-07-07
fix(ci): EC2 staging deploy silently skipped migrate + restart

Bahugunajii:
## Symptom
On the EC2 staging deploy (`staging_deploy.yml`), the `.env` looked un-applied and `mix ecto.migrate` (plus swagger/assets and the **server restart**) never ran — yet the job showed **green**.

## Root cause
The remote script is piped into `bash -se` over the ssh **stdin** (`ssh ... "bash -se" <<'HEREDOC'`). `bash -s` reads its *script* from stdin. A command after `mix deps.get` (a Hex/rebar prompt, or a `sudo` password prompt) read from that same stdin and **swallowed the remaining script lines**. bash then hit EOF and exited **0**, so `deps.compile`, `ecto.migrate`, the asset build, and the server restart were all skipped — and the job still passed.

Confirmed from the run log: output stops right after the `deps.get` dependency list — no "Compiling", no "Migrations", and not even the final `echo` — then straight to job cleanup, exit 0. The `.env` *was* patched (`config/.env` on the host was rewritten at deploy time), but the app was never restarted to pick it up and migrations never ran.

Note: `config/.env` is **untracked**, so `git checkout .` does *not* revert it — the checkout ordering is not the problem.

## Fixes
- **Wrap the remote body in a brace group with stdin from `/dev/null`** — bash parses the whole group before executing it, and its stdin is `/dev/null`, so no inner command can consume the script. Core fix.
- **Health-check `/api/health`** (public, and it verifies the DB) instead of `/api/session?limit=1` (which requires a bearer token and would 401), and **exit non-zero on non-200** so a broken boot fails the job instead of reporting success.
- **Guard that `config/.env` exists** before the `sed` injection, with a clear error otherwise.

## Validation
`actionlint` reports no issues on the changed step. The two warnings it flags are pre-existing on the untouched "Determine branch to deploy" step (`SC2086` on `$GITHUB_OUTPUT` and the `github.head_ref` injection note) — out of scope here; happy to clean those separately.

⚠️ This can only be exercised end-to-end by an actual deploy — recommend a `workflow_dispatch` run of this branch to confirm migrate + restart now run and the health check gates correctly.

### DIFF
--- a/.github/workflows/staging_deploy.yml
+++ b/.github/workflows/staging_deploy.yml
@@ -79,7 +79,7 @@ jobs:
             sudo env MIX_ENV=prod mix assets.deploy
             sudo fuser -k "${APP_PORT}/tcp" 2>/dev/null || true
             sleep 2
-            nohup sudo env MIX_ENV=prod elixir --erl \"-detached\" -S mix phx.server > /tmp/phx_server.log 2>&1 &
+            nohup sudo env MIX_ENV=prod mix phx.server > /tmp/phx_server.log 2>&1 &
             # Hit /api/health (public, no bearer token, and it verifies the DB) —
             # /api/session requires auth and would 401. Poll for up to ~2 min: a
             # cold prod boot (release load + DB pool) can take well over 5s. Fail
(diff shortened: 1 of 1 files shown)


Bahugunajii:
Closing — obsolete after the staging cutover to ECS Fargate (2026-07-12). staging-db.avantifellows.org now resolves to the ALB/ECS service, and the EC2 SSH-based `staging_deploy.yml` has been disabled (renamed `.disabled`) in #586. No longer deploying to the EC2 box, so these fixes aren't needed.

PR #589, opened 2026-07-09
perf(session-occurrence): order active-window queries by end_time, not id

### LINKED ISSUE #491: Replace ORDER BY id with temporal ordering for session_occurrence active-window queries
## What's happening now

The `session_occurrence` controller uses `ORDER BY id` as its default pagination ordering for all queries, including temporal "active session" lookups. When a client requests active sessions (sessions where `start_time <= now AND end_time >= now`), PostgreSQL chooses to walk the primary key index in sequence, scanning hundreds of thousands of rows to find the small number of currently-active sessions.

The query:

```sql
SELECT ... FROM session_occurrence
WHERE start_time <= $1 AND end_time >= $2
ORDER BY id
LIMIT 10
```

The cause is in `lib/dbservice_web/controllers/session_occurence_controller.ex:69` — the `base_query` always applies `order_by: [asc: m.id]`. The `active` filter at line 205 then adds the temporal `WHERE` clause on top of this ordering.

**Audit reference:** `db-investigation.md`, Section 2.3a (F11), Tier 1 Action #1

**Confirmed by:** Codex (GPT-5.5), Gemini — both rate Critical

---

### PostgreSQL makes a bad plan choice because of the ORDER BY

When the query says `ORDER BY id LIMIT 10`, the planner has two options:
1. Use the `end_time` index to find matching rows quickly, then sort them by `id` — fast filtering but requires a sort step
2. Walk the primary key index (`id`) in order, checking each row against the time filter — no sort needed, but must scan until it finds 10 matches

PostgreSQL picks option 2 because it avoids a sort. The problem: active sessions are a tiny fraction of the ~800K rows in the table. The planner ends up scanning 630,180 rows (reading 618,673 buffer pages — ~4.8 GB of logical I/O) just to find 10 matching rows.

**EXPLAIN ANALYZE — current (ORDER BY id):**
```
Limit (actual time=255.682..256.660 rows=10)
  Index Scan using session_occurence_pkey
    Filter: (start_time <= X AND end_time >= X)
    Rows Removed by Filter: 630,180
    Buffers: shared hit=618,673
Execution Time: 256.699 ms
```

**EXPLAIN ANALYZE — with ORDER BY end_time:**
```
Index Scan using session_occurrence_end_time_index
  Index Cond: (end_time >= X)
  Filter: (start_time <= X)
  Rows Removed by Filter: 66
Execution Time: 0.455 ms
```

The difference: **565x faster** (0.455ms vs 256.699ms). Instead of scanning 630K rows, it scans 76 rows. This is the single highest single-query impact fix in the entire database.

---

### Step 1 — Change the ORDER BY for the active-window query path

In `lib/dbservice_web/controllers/session_occurence_controller.ex`, modify the active-window query to use temporal ordering with an `id` tie-breaker:

```elixir
# Instead of the base_query's order_by: [asc: m.id]
# For the active filter path, use:
order_by: [asc: so.end_time, asc: so.id]
```

Breaking this down:
- `end_time` as the primary sort key — PostgreSQL will use the existing `session_occurrence_end_time_index` to jump directly to rows where `end_time >= now`, then filter on `start_time <= now`. Only ~66 rows need to be checked instead of 630K.
- `id` as the secondary sort key — this is a **tie-breaker for deterministic pagination**. Sessions frequently share the same `end_time` (e.g., classes ending at the top of the hour). Without a tie-breaker, rows with identical `end_time` values can appear in inconsistent order across pages, causing duplicates or missing records when paginating.

**Do not change the default `ORDER BY id` for non-temporal queries.** The fix should be scoped specifically to the active-window filter path (`is_start_time=active`), not applied globally to all session_occurrence queries.

### Step 2 — Handle the API ordering change

This change alters result ordering from "insertion order" to "ending soonest first." If API consumers depend on the current ordering:

- Check if any frontend or external consumer relies on `session_occurrence` results being sorted by `id`
- If consumers exist that depend on ordering, consider adding an explicit `sort_by` query parameter (e.g., `?sort_by=end_time`) and rolling it out as opt-in before changing the default
- If no consumers depend on specific ordering (likely — the `active` filter is about "what's happening now," not historical browsing), the change can be applied directly

### Step 3 — Optional: Add a composite index for perfect keyset pagination

For future keyset/cursor pagination support (which eliminates the performance problems of deep OFFSET pagination entirely):

```sql
CREATE INDEX CONCURRENTLY session_occurrence_end_time_id_index
ON session_occurrence (end_time, id);
```

This index supports both the temporal filter and the tie-breaking sort in a single B-tree traversal. Not strictly necessary for the immediate fix (the existing `end_time` index handles it), but becomes valuable if the API adopts cursor-based pagination.

---

## What NOT to do

- **Don't use `ORDER BY end_time` without a tie-breaker.** Ties in `end_time` cause non-deterministic pagination — rows shift between pages on repeated queries.
- **Don't change the global `base_query` ordering.** Other query paths (listing all occurrences, filtering by session_id) may work fine with `ORDER BY id`. Only the temporal active-window path needs this fix.
- **Don't use OFFSET for deep pagination.** OFFSET-based pagination degrades linearly with depth. If users frequently paginate past page 10, consider switching to keyset pagination using `(end_time, id)` as the cursor.

---

## Severity

**Critical** — confirmed by both council reviewers. This is a 565x speedup on a production query that scans 630K rows and reads ~4.8 GB of logical I/O per execution. It is the single highest single-query impact fix identified in the database audit.

Bahugunajii:
Closes #491

## Why

The active-window lookup `WHERE start_time <= now AND end_time >= now` inherited the controller's base `ORDER BY id`. With `ORDER BY id LIMIT n`, PostgreSQL walks the primary-key index and scans ~630K rows to find the handful of currently-active occurrences (~256ms, ~4.8 GB logical I/O per call). Ordering by `end_time` instead lets it use the `end_time` index and check ~66 rows (~0.45ms) — a ~565x speedup (the audit's highest single-query impact).

## What

- **Scoped to the `active` path only** (base `ORDER BY id` is unchanged for all other query paths, per the issue): `exclude(:order_by)` the inherited id ordering, then `ORDER BY end_time, id`. `id` is a deterministic pagination tie-breaker for occurrences that share an `end_time` (common — classes ending on the hour), preventing rows from shifting between pages.
- **Composite `(end_time, id)` index** (built `CONCURRENTLY`) so the temporal filter and the tie-breaker are served in a single B-tree traversal and future keyset/cursor pagination is supported. (The existing `end_time` index already delivers the speedup; this makes the ordering fully index-backed.)
- Controller test asserting `is_start_time=active` returns occurrences in `end_time`-ascending order regardless of insertion order.

## Behavior note
This changes the ordering of `active` results from insertion order (`id`) to "ending soonest first" (`end_time`). The `active` filter is a "what's happening now" view, so this is expected/desirable; other list paths are unaffected.

## Verification
`mix test test/dbservice_web/controllers/session_occurence_controller_test.exs` → 7/7; credo + format clean.

### DIFF
--- a/lib/dbservice_web/controllers/session_occurence_controller.ex
+++ b/lib/dbservice_web/controllers/session_occurence_controller.ex
@@ -203,9 +203,14 @@ defmodule DbserviceWeb.SessionOccurrenceController do
             )
 
       "active" ->
-        from(so in query,
-          where: so.start_time <= ^current_time and so.end_time >= ^current_time
-        )
+        # Active-window lookups filter by end_time (served by the end_time index). Override
+        # the base `ORDER BY id` - which makes PostgreSQL walk the primary key and scan
+        # ~630K rows to find the few active ones - with temporal ordering. `id` is a
+        # deterministic pagination tie-breaker for rows sharing an end_time.
+        query
+        |> exclude(:order_by)
+        |> where([so], so.start_time <= ^current_time and so.end_time >= ^current_time)
+        |> order_by([so], asc: so.end_time, asc: so.id)
 
       _ ->
         query
--- a/priv/repo/migrations/20260709120000_add_session_occurrence_end_time_id_index.exs
+++ b/priv/repo/migrations/20260709120000_add_session_occurrence_end_time_id_index.exs
@@ -1,0 +1,14 @@
+defmodule Dbservice.Repo.Migrations.AddSessionOccurrenceEndTimeIdIndex do
+  use Ecto.Migration
+
+  @disable_ddl_transaction true
+  @disable_migration_lock true
+
+  # Backs the active-window ordering `ORDER BY end_time, id`: the existing single-column
+  # end_time index already serves the temporal filter, but a composite (end_time, id) lets
+  # PostgreSQL satisfy the filter and the id tie-breaker in one B-tree traversal (and
+  # supports future keyset/cursor pagination on that same key).
+  def change do
+    create_if_not_exists index(:session_occurrence, [:end_time, :id], concurrently: true)
+  end
+end
--- a/test/dbservice_web/controllers/session_occurence_controller_test.exs
+++ b/test/dbservice_web/controllers/session_occurence_controller_test.exs
@@ -39,13 +39,15 @@ defmodule DbserviceWeb.SessionOccurrenceControllerTest do
     test "is_start_time=active returns active occurrences ordered by end_time ascending", %{
       conn: conn
     } do
-      past = ~U[2020-01-01 00:00:00Z]
+      now = DateTime.utc_now()
+      past = DateTime.add(now, -365, :day)
 
       # All currently active (started in the past, end in the future), inserted in an order
-      # that differs from their end_time order.
-      mid = session_occurrence_fixture(%{start_time: past, end_time: ~U[2030-06-01 00:00:00Z]})
-      late = session_occurrence_fixture(%{start_time: past, end_time: ~U[2031-06-01 00:00:00Z]})
-      soon = session_occurrence_fixture(%{start_time: past, end_time: ~U[2029-06-01 00:00:00Z]})
+      # that differs from their end_time order. Anchored to `now` with fixed offsets so the
+      # relative ordering stays deterministic and the fixtures never expire with the calendar.
+      mid = session_occurrence_fixture(%{start_time: past, end_time: DateTime.add(now, 2, :day)})
+      late = session_occurrence_fixture(%{start_time: past, end_time: DateTime.add(now, 3, :day)})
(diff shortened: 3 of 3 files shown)


Bahugunajii:
Applied the concurrent-index review guidance from #569/#570 to the `(end_time, id)` migration: `create` instead of `create_if_not_exists`, removed `@disable_migration_lock true` (kept `@disable_ddl_transaction true`), and set `migration_lock: :pg_advisory_lock` on the repo. App-code change is unchanged; `session_occurence_controller_test` still 7/7, and the index builds `indisvalid = true` locally.

Note: the one-line `migration_lock` config also appears in #570/#571/#572 — trivial merge conflict (keep one copy).

deepansh96 (approved):

PR #595, opened 2026-07-10
feat(session-occurrence): add start_time_lte/end_time_gte window filters

Bahugunajii:
## Problem
A **continuous** session (one occurrence spanning e.g. 9 Jul 2pm–6pm) never shows on the Gurukul home page before it opens — no "Starts at 2:00 PM" state. Gurukul queries `POST /session-occurrence/search` with `is_start_time=today`, whose continuous branch only returns occurrences that are *currently active* (`start_time <= now AND end_time >= now`), so a not-yet-started occurrence is never returned.

## Design
Per discussion (Pritam): db-service should not encode what Gurukul wants to show. It should just answer "which occurrences are active within a given time window" and let the client own visible/starts-soon logic.

## Change
Two new generic params on `GET /session-occurrence` and `POST /session-occurrence/search` (naming follows the existing `end_time_gte`/`end_time_lt` filters on `/session/search`):

- `start_time_lte` — occurrence starts at or before this ISO-8601 timestamp
- `end_time_gte` — occurrence ends at or after this ISO-8601 timestamp

Together they select occurrences overlapping the window `[end_time_gte, start_time_lte]`; each also works alone. Invalid timestamps return `400` (parsed with the existing `Util.parse_datetime`).

**Backward compatible:** `is_start_time=today|active` behavior is untouched; existing callers need no changes. Gurukul will switch to sending `end_time_gte=<now>` and `start_time_lte=<end of today>` in a follow-up PR (deploy this first).

## Tests
- Window overlap via `POST /search`: upcoming-today returned (the bug scenario), already-ended excluded, starts-tomorrow excluded, multi-day-still-open returned
- Each param alone
- Invalid timestamp → 400

`mix test`: the 4 new tests pass; the suite's 8 pre-existing failures are identical on clean `main` (unrelated: student import + error JSON wording).

### DIFF
--- a/lib/dbservice_web/controllers/session_occurence_controller.ex
+++ b/lib/dbservice_web/controllers/session_occurence_controller.ex
@@ -5,6 +5,7 @@ defmodule DbserviceWeb.SessionOccurrenceController do
   alias Dbservice.Repo
   alias Dbservice.Sessions
   alias Dbservice.Sessions.SessionOccurrence
+  alias Dbservice.Utils.Util
 
   action_fallback(DbserviceWeb.FallbackController)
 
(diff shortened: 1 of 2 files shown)


suryabulusu (approved):
lgtm

Issue #602, opened 2026-07-14
Support multi-language problem creation via lang_versions in POST /resource


ankurcdoshi:
## Context

The CMS frontend is being updated to support creating problems in multiple languages (en, hi, gu, ta) in a single API call. The `POST /resource` endpoint needs to accept a new `lang_versions` array while remaining backward compatible with the existing frontend which sends the old single-language format.

---

### `POST /resource` — Create Problem

Support both old and new request body formats based on presence of `lang_versions`.

**Logic:**
- If `lang_versions` is present and non-empty → use new multi-lang path: insert one row in `problem_lang` per entry in `lang_versions`
- If `lang_versions` is absent/null → use existing single-lang path unchanged (**Can remove this once new code is deployed from frontend**)

---

### Old format (existing frontend — must keep working)

```json
{
  "lang_code": "en",
  "type": "problem",
  "subtype": "mcq_single_answer",
  "skill_ids": [1, 2],
  "type_params": { "test_ids": [] },
  "curriculum_grades": [{ "curriculum_id": 123, "grade_id": 8 }],
  "subject_id": 5,
  "topic_id": 42,
  "chapter_id": 10,
  "difficulty_level": "medium",
  "concept_ids": [5],
  "tags": ["algebra"],
  "meta_data": {
    "text": "<p>What is 2+2?</p>",
    "options": ["<p>3</p>", "<p>4</p>", "<p>5</p>", "<p>6</p>"],
    "answer": ["2"],
    "solutions": [{ "type": "text", "value": "<p>2+2=4</p>" }]
  }
}
```

---

### New format (updated frontend)

```json
{
  "type": "problem",
  "subtype": "mcq_single_answer",
  "skill_ids": [1, 2],
  "type_params": { "test_ids": [] },
  "curriculum_grades": [{ "curriculum_id": 123, "grade_id": 8 }],
  "subject_id": 5,
  "topic_id": 42,
  "chapter_id": 10,
  "difficulty_level": "medium",
  "concept_ids": [5],
  "tags": ["algebra"],
  "lang_versions": [
    {
      "lang_code": "en",
      "meta_data": {
        "text": "<p>What is 2+2?</p>",
        "options": ["<p>3</p>", "<p>4</p>", "<p>5</p>", "<p>6</p>"],
        "answer": ["2"],
        "solutions": [{ "type": "text", "value": "<p>2+2=4</p>" }]
      }
    },
    {
      "lang_code": "hi",
      "meta_data": {
[...]

PR #612, opened 2026-07-15
feat(api): return lang_versions in all problem GET responses

### LINKED ISSUE #610: Standardize problem response format to return `lang_versions` across all GET APIs
## Context

As part of multi-language support, the problem response format needs to be standardized across all GET APIs to always return a `lang_versions` array instead of a single flat `meta_data` object. This ensures a consistent shape regardless of which endpoint returns problem data.

## APIs to update

- `GET /resource/problem/{id}`
- `GET /resource/test/{id}/problems`
- `GET /problems`
- `GET /problems/search`

## Backward Compatibility

Both `meta_data` (old) and `lang_versions` (new) must be returned together during the transition period, until the new frontend is deployed. This is the same pattern as the POST side in #602.

- Currently deployed frontend keeps working — reads `meta_data` as before, ignores `lang_versions`
- New frontend reads `lang_versions`, ignores top-level `meta_data`
- Once new frontend is deployed, `meta_data` at root level can be deprecated and removed in a follow-up cleanup

## Required response shape change for following fields (Remaining fields in response should be kept as it is)

```json
{
  "lang_code": "en",
  "meta_data": {
    "text": "What is 2+2?",
    "options": ["1", "2", "3", "4"],
    "answer": ["3"],
    "solutions": [{ "type": "text", "value": "Simple addition." }]
  },
  "lang_versions": [
    {
      "lang_code": "en",
      "meta_data": {
        "text": "What is 2+2?",
        "options": ["1", "2", "3", "4"],
        "answer": ["3"],
        "solutions": [{ "type": "text", "value": "Simple addition." }]
      }
    },
    {
      "lang_code": "hi",
      "meta_data": {
        "text": "2+2 क्या है?",
        "options": ["१", "२", "३", "४"],
        "answer": ["3"],
        "solutions": [{ "type": "text", "value": "सरल जोड़।" }]
      }
    }
  ]
}
```

Bahugunajii:
Closes #610

## What

Standardizes the problem response format across all problem GET APIs to include a `lang_versions` array — one `{lang_code, meta_data}` entry per `problem_lang` row of the resource:

- `GET /api/resource/problem/{problem_id}/{lang_code}/{curriculum_id}`
- `GET /api/resource/test/{id}/problems`
- `GET /api/problems`
- `GET /api/problems/search`

## Backward compatibility

Both shapes are returned together during the transition (same pattern as the POST side in #602):

- Top-level `meta_data` (and `lang_code` where present) is unchanged and still reflects the requested language — the currently deployed frontend keeps working.
- `lang_versions` carries every available language for the new frontend.
- Once the new frontend is deployed, top-level `meta_data` can be removed in a follow-up cleanup.

Remaining response fields are untouched.

## How

- New `ProblemLanguages.list_lang_versions_by_resource_id/1` builds the array from `problem_lang` joined with `language`, ordered by insertion.
- `ResourceJSON.render_problem/1` (serves `/problems`, `/problems/search`, `/resource/test/:id/problems`) and `ResourceJSON.problem_lang/1` (serves `/resource/problem/...`, and the PATCH problem response) now include `lang_versions`.
- Swagger `ProblemResource` schema documents the new field.

## Testing

- New tests for the context function and both JSON render paths (`mix test` green; `mix format --check-formatted` and `mix credo` clean).
- Verified live against the dev server on all four endpoints with a problem having `en` + `hi` rows, e.g. `GET /api/resource/problem/503/en/1` returned the old flat `meta_data` (en) plus `lang_versions: [{en, ...}, {hi, ...}]`; single-language problems return a one-entry array.

### DIFF
--- a/lib/dbservice_web/json/resource_json.ex
+++ b/lib/dbservice_web/json/resource_json.ex
@@ -4,6 +4,7 @@ defmodule DbserviceWeb.ResourceJSON do
   alias Dbservice.Resources.ResourceChapter
   alias Dbservice.Concepts
   alias Dbservice.Paragraphs
+  alias Dbservice.ProblemLanguages
   alias Dbservice.ResourceConcepts
   alias Dbservice.ResourceCurriculums
   alias Dbservice.Repo
@@ -193,12 +194,15 @@ defmodule DbserviceWeb.ResourceJSON do
         subject_id: Map.get(curriculum, :subject_id, nil)
       })
 
-    # Add problem language data
+    # Add problem language data. Top-level meta_data keeps the currently
+    # requested language for backward compatibility (see issue #610);
+    # lang_versions carries every available language for the new frontend.
     problem_map =
       Map.merge(resource_with_curriculum, %{
         meta_data: Map.get(problem_lang, :meta_data, nil),
         lang_id: Map.get(problem_lang, :lang_id, nil),
-        paragraph_id: Map.get(problem_lang, :paragraph_id, nil)
+        paragraph_id: Map.get(problem_lang, :paragraph_id, nil),
+        lang_versions: ProblemLanguages.list_lang_versions_by_resource_id(resource.id)
       })
 
     # Fetch concept information
@@ -239,8 +243,12 @@ defmodule DbserviceWeb.ResourceJSON do
     # top-level fields (merged below from the resolved resource_curriculum), so
     # the curriculum_grades list is redundant here.
     |> Map.delete(:curriculum_grades)
+    # Top-level meta_data/lang_code keep the requested language for backward
+    # compatibility (see issue #610); lang_versions carries every available
+    # language for the new frontend.
     |> Map.put(:meta_data, meta_data)
     |> Map.put(:lang_code, lang_code)
+    |> Map.put(:lang_versions, ProblemLanguages.list_lang_versions_by_resource_id(resource.id))
     |> Map.merge(%{
       curriculum_id: rc.curriculum_id,
       grade_id: rc.grade_id,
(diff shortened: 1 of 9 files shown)


lib/dbservice_web/json/resource_json.ex
```diff
@@ -193,12 +194,15 @@ defmodule DbserviceWeb.ResourceJSON do
...
+    # Add problem language data. Top-level meta_data keeps the currently
+    # requested language for backward compatibility (see issue #610);
+    # lang_versions carries every available language for the new frontend.
     problem_map =
       Map.merge(resource_with_curriculum, %{
         meta_data: Map.get(problem_lang, :meta_data, nil),
         lang_id: Map.get(problem_lang, :lang_id, nil),
-        paragraph_id: Map.get(problem_lang, :paragraph_id, nil)
+        paragraph_id: Map.get(problem_lang, :paragraph_id, nil),
+        lang_versions: ProblemLanguages.list_lang_versions_by_resource_id(resource.id)
```
ankurcdoshi: `list_lang_versions_by_resource_id/1` is called per problem inside `problems/1` which renders a list. If a test has 50 problems, that's 50 additional DB queries. Can we either:

1. Batch fetch all lang_versions for all problem IDs at once and group in Elixir? or
2. Use a join/preload in the original problems query?

ankurcdoshi (changes requested):
Just noticed that right now `lang_code` is required in following apis. Can we create new apis which can work without `lang_code`? Because for each problem we need data in all languages, and language specific filtering I can apply at front end.
1. `GET /resource/problem/{id}/{lang_code}/{curriculum_id}`
2. `GET /resource/test/{id}/problems?lang_code={lang_code}&curriculum_id={curriculum_id}`
3. `GET /api/problems?topic_id={topic_id}&curriculum_id={curriculum_id}&lang_code={lang_code}`
4. `GET /problems/search?limit={limit}&search={search_query}&offset={offset}&lang_code={lang_code}`

_Note_: Keep existing api for backward compatibility right now, which can be removed later on.

ankurcdoshi (approved):

PR #613, opened 2026-07-15
feat(api): accept lang_versions in problem create endpoints (POST /resource + batch)

Bahugunajii:
Closes #602
Closes #603

## What

Both problem-create endpoints now accept the new multi-language format:

- `POST /api/resource` (single problem — #602)
- `POST /api/resources/problems/batch` (comprehension batch, the endpoint behind the frontend's create-problems flow — #603)

When `lang_versions` is present and non-empty, one `problem_lang` row is inserted per entry (`lang_code` resolved to `lang_id`, `meta_data` stored per language). Both endpoints share the same insertion path (`insert_problem_language`), so the batch endpoint gets the behavior from the same code.

## Backward compatibility

- If `lang_versions` is absent/null, the existing flat `lang_code` + `meta_data` path runs unchanged — the currently deployed frontend keeps working.
- If both are present, `lang_versions` wins.
- The fallback can be removed in a follow-up once the new frontend is deployed.

## Details

- **Comprehension + multiple languages**: the `paragraph` body is created **once** and its id shared by every `problem_lang` row of the problem (previously the per-language insert would have created a duplicate paragraph per language). The batch endpoint's shared top-level `paragraph` continues to work — the injected `paragraph_id` is reused by all lang_versions rows.
- **Unknown `lang_code`** inside `lang_versions` rolls back the whole resource create and returns 422 with a clear error (per-index failure in the batch endpoint, other entries still succeed). The old flat path keeps its existing silently-skip behavior untouched.
- Swagger: `Resource` request schema documents `lang_versions`; batch request example shows both formats.

## Testing

- 7 new controller (ConnCase) tests covering: multi-language create, old-format fallback, precedence, unknown-language rollback, comprehension paragraph sharing across languages, batch create with lang_versions + shared paragraph, and batch mixed old/new with per-index failure.
- `mix test`, `mix format --check-formatted`, `mix credo` all green.

## Related

- GET-side counterpart (response `lang_versions`): #612 (issue #610)
- PATCH-side counterpart (#611) coming as a separate PR.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #603, opened 2026-07-15
Support `lang_versions` in POST /create-problems (comprehension type)


ankurcdoshi:
### Context

Issue #602 covers `lang_versions` support for `POST /resource` (single problem). Comprehension problems use a separate endpoint — `POST /create-problems` — which sends a paragraph and an array of problems in a single call. This issue tracks the equivalent change for that endpoint.

Note: Comprehension type does **not** support multiple languages right now (planned for future). Every problem inside `problems` will always have a single entry in `lang_versions` with `lang_code: "en"`. The only change needed is that the old flat `lang_code` + `meta_data` at the problem root is replaced by a `lang_versions` array, consistent with what #602 defines for `POST /resource`.

---

### Old payload structure

```json
{
  "paragraph": "<p>Read the following passage...</p>",
  "problems": [
    {
      "type": "problem",
      "subtype": "comprehension",
      "skill_ids": [1],
      "difficulty_level": "easy",
      "topic_id": 10,
      "chapter_id": 2,
      "subject_id": 1,
      "curriculum_grades": [{ "curriculum_id": 1, "grade_id": 5 }],
      "type_params": { "test_ids": [] },
      "concept_ids": [3],
      "tags": [],
      "lang_code": "en",
      "meta_data": {
        "text": "What year did it happen?",
        "options": null,
        "answer": ["1947"],
        "solutions": [{ "type": "text", "value": "It happened in 1947." }]
      }
    }
  ]
}
```

### New payload structure

```json
{
  "paragraph": "<p>Read the following passage...</p>",
  "problems": [
    {
      "type": "problem",
      "subtype": "comprehension",
      "skill_ids": [1],
      "difficulty_level": "easy",
      "topic_id": 10,
      "chapter_id": 2,
      "subject_id": 1,
      "curriculum_grades": [{ "curriculum_id": 1, "grade_id": 5 }],
      "type_params": { "test_ids": [] },
      "concept_ids": [3],
      "tags": [],
      "lang_versions": [
        {
          "lang_code": "en",
          "meta_data": {
            "text": "What year did it happen?",
            "options": null,
            "answer": ["1947"],
            "solutions": [{ "type": "text", "value": "It happened in 1947." }]
          }
        }
      ]
    }
  ]
}
```

---

### What needs to change

Each problem object inside `problems` should be processed using the same `lang_versions` logic defined in #602, including the same backward compatibility requirement:

- If `lang_versions` is present and non-empty → use new multi-lang path
[...]

PR #614, opened 2026-07-15
feat(api): sync problem_lang rows from lang_versions in PATCH /resource

### LINKED ISSUE #611: Support `lang_versions` in PATCH /resource/{id} for multi-language problem updates
## Context

As part of multi-language support (#602 covers POST), the PATCH endpoint for updating a problem needs to handle `lang_versions` similarly. However, unlike POST, PATCH has an additional concern — if a language that was previously saved is no longer included in the request, it should be removed from the database.

## How the frontend sends data

The frontend only includes languages the user has explicitly enabled via the language tab checkboxes. So if a problem was previously saved with English + Hindi, but the user unchecks Hindi and saves, the PATCH payload will only contain English in `lang_versions`. The backend must treat missing languages as deletions.

## Required behavior

Given the `lang_versions` array in the PATCH request, the backend should:

1. **Upsert** — for each language present in `lang_versions`, insert or update its row in `problem_lang` table
2. **Delete** — for each language that exists in the DB for this problem but is absent from the incoming `lang_versions`, delete that row from `problem_lang` table

## Backward compatibility

Same pattern as #602 — if `lang_versions` is absent/null in the request, fall back to existing single-lang path using top-level `lang_code` + `meta_data` unchanged.

Once the new frontend is deployed, the backward compatibility fallback (old flat `lang_code` + `meta_data` support) should be removed in a follow-up cleanup, same as planned in #602.

## Payload shape

```json
{
  "lang_versions": [
    {
      "lang_code": "en",
      "meta_data": {
        "text": "What is 2+2?",
        "options": ["1", "2", "3", "4"],
        "answer": ["3"],
        "solutions": [{ "type": "text", "value": "Simple addition." }]
      }
    }
  ]
}
```

In the above example, if Hindi was previously saved for this problem, it should be deleted since it is not present in `lang_versions`.

Bahugunajii:
Closes #611

## What

`PATCH /api/resource/{id}` now accepts the `lang_versions` array and syncs the problem's `problem_lang` rows to **exactly** the languages in the request:

1. **Upsert** — each `{lang_code, meta_data}` entry inserts a new row or updates the existing one for that language
2. **Delete** — languages present in the DB but absent from `lang_versions` are deleted (the frontend only sends languages the user has enabled via the language tabs, so missing = unchecked)

`PATCH /api/resources/problems/batch` shares the same update path (`update_resource_and_associations`), so batch updates get the behavior too.

## Backward compatibility

Same pattern as #602/#613 — when `lang_versions` is absent/null, the existing flat `lang_code` + `meta_data` path runs unchanged (which only updates the addressed language and never deletes). `lang_versions` takes precedence when present. The fallback can be removed once the new frontend is deployed.

## Details

- **Comprehension problems**: a language added via `lang_versions` joins the problem's existing shared paragraph (its `paragraph_id` is inherited), and a `paragraph` body in the same request still updates the shared paragraph exactly once via the existing pre-sync.
- **Unknown `lang_code`** entries are skipped without affecting valid entries, matching the flat path's no-op behavior on unknown languages (PATCH association updates run outside a transaction, so there is no rollback to lean on).
- Swagger request docs for `lang_versions` land in #613; no schema changes here to avoid conflicts.

## Testing

- 6 new DataCase tests: upsert (update + insert), delete-on-absence, unknown-language skip, old-flat-format unchanged (no deletions), comprehension paragraph inheritance, and shared-paragraph body update.
- Verified live against the dev server: `PATCH /api/resource/503` with `[en, hi]` inserted the Hindi row; a follow-up PATCH with `[en]` deleted it. DB restored to its original state afterwards.
- `mix test`, `mix format --check-formatted`, `mix credo` all green.

## Related

- POST side (#602, #603): #613
- GET side (#610): #612

### DIFF
--- a/lib/dbservice/resources.ex
+++ b/lib/dbservice/resources.ex
@@ -937,6 +937,10 @@ defmodule Dbservice.Resources do
 
   defp update_problem_language(resource, params) do
     cond do
+      has_lang_versions?(params) ->
+        params = pre_sync_shared_paragraph(resource, params)
+        sync_problem_lang_versions(resource, params)
+
       has_lang_code?(params) ->
         params = pre_sync_shared_paragraph(resource, params)
         update_problem_language_for_lang(resource, params)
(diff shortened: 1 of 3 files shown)


ankurcdoshi (changes requested):
Need `lang_versions` in response as well

ankurcdoshi (approved):

PR #632, opened 2026-07-17
fix(api): enforce default and max pagination limits on list endpoints

Bahugunajii:
## Context

Prod went down today (2026-07-17) when a bot hit `GET /api/student?offset=0&limit=100000` and the Phoenix service hung trying to serve it. List endpoints passed the raw `limit`/`offset` query params straight into Ecto queries — no cap, and **no limit at all** when the param was omitted.

## Change

New `Dbservice.Utils.Pagination` module, applied to **all 52 controllers** that accept `offset`/`limit`:

| Input | Behavior |
|---|---|
| No `limit` param | default limit **1000** |
| `limit=100000` | clamped to max **10000** |
| `limit=0`, `limit=-5` | clamped to 1 |
| `limit=abc` (previously a 500) | falls back to default |
| `offset` missing / negative / garbage | 0 |

Limits are configurable in one place:

```elixir
config :dbservice, Dbservice.Utils.Pagination,
  default_limit: 1000,
  max_limit: 10_000
```

Also fixes `ResourceController.apply_pagination/2`, which skipped pagination entirely when `limit` was absent.

## ⚠️ Behavior change for consumers

Any client calling a list endpoint **without a `limit` param currently gets every row — after this it gets at most 1000**. Clients that genuinely need everything must paginate with `offset`/`limit` (max 10000 per page). If any known consumer (portal, Gurukul, ETL jobs) relies on fetch-all behavior, we should bump `default_limit` or fix that consumer before merging.

## Testing

- `mix test test/dbservice/utils/pagination_test.exs` — 8 new tests, all green
- `mix compile --warnings-as-errors`, `mix credo`, `mix format --check-formatted` all clean
- Verified generated SQL via `Repo.to_sql`: `limit=100000` → `LIMIT 10000`; no param → `LIMIT 1000`; `limit=50&offset=100` passes through untouched
- Full `mix test`: 8 failures, all pre-existing on main (enrollment service / group update processor / student create status codes — unrelated to this change)

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -10,6 +10,12 @@ import Config
 config :dbservice,
   ecto_repos: [Dbservice.Repo]
 
+# Pagination guardrails for list endpoints: applied when no `limit` query
+# param is given (default_limit) and as a hard cap when one is (max_limit).
+config :dbservice, Dbservice.Utils.Pagination,
+  default_limit: 1000,
+  max_limit: 10_000
+
 # Configures the endpoint
 config :dbservice, DbserviceWeb.Endpoint,
   load_from_system_env: false,
--- a/lib/dbservice/resources.ex
+++ b/lib/dbservice/resources.ex
@@ -12,6 +12,7 @@ defmodule Dbservice.Resources do
   alias Dbservice.Languages.Language
   alias Dbservice.Resources.{ResourceTopic, ResourceChapter, ResourceConcept}
   alias Dbservice.Utils.Util
+  alias Dbservice.Utils.Pagination
   alias Dbservice.CmsStatuses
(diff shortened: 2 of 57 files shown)


deepansh96:
Two endpoints still bypass the new pagination limits:

- `GET /api/resource`
- `GET /api/problems/search`

Both use the raw `limit` and `offset` values in `Dbservice.Resources`, so requests like `?limit=100000` are still possible. The problem search endpoint is especially risky because it runs extra database queries for every result.

Please use `Pagination.limit(params)` and `Pagination.offset(params)` in both pagination functions in `Dbservice.Resources`, and add a test that checks these endpoints are capped.

pritamps (approved):

PR #638, opened 2026-07-20
Add student_enrollment import type to backfill enrollments for existing students

Bahugunajii:
## Problem

We need to backfill historical enrollment data for students whose **student records already exist** in the DB but who have **no group-user and no enrollment (ER) records** (auth_group / school / batch / grade memberships).

The existing `student` import can't do this — when a student already exists, `process_student_record/1` deliberately short-circuits with *"Student already exists ... Use 'Update Student Details' import type for updates."* And `student_update` only patches profile fields; it never creates group-user/ER records. So there was no bulk path to create enrollments for an already-existing student.

## Solution

Add a new, non-breaking import type **`student_enrollment` ("Enroll Existing Students")** that:
- looks up the existing student by `student_id` / `apaar_id`,
- errors strictly with *"Student not found"* if the student doesn't exist (these students are expected to already exist — a miss signals a data problem worth surfacing),
- **hard-errors if the student is already mapped to any of the row's memberships** (see "Strict already-enrolled guard" below),
- otherwise reuses the existing `StudentEnrollment.create_enrollments/2` logic to create the auth_group/school/batch/grade group-user + ER records.

## Strict already-enrolled guard

`create_enrollments/2` is *idempotent* by design: for an identical existing mapping it silently no-ops, and it only errors when the student is enrolled in a *different* exclusive group (school/grade/auth_group). For this backfill that's too lenient — an already-enrolled student would be silently skipped instead of surfaced.

So before calling `create_enrollments/2`, the import now runs a strict pre-check, `EnrollmentService.validate_row_memberships_absent/2`:

- **Per-membership**: for each of `auth_group` / `school` / `batch` / `grade` named in the row, it independently checks whether the student is already mapped to *that exact group*.
- **Both tables**: it hard-errors if a `group_user` row exists **or** a **current** (`is_current: true`) enrollment record exists for that group.
- **Current ERs only**: closed/historical ERs (`is_current: false`) are ignored, so a previously-dropped-out student can still be backfilled.
- Blank/absent identifiers are skipped; an unresolvable school/batch/grade/auth_group surfaces its normal "not found" error.

This also closes the group_user/ER drift gap for the row's own memberships: if a `group_user` is missing but a current ER exists (or vice-versa), the guard catches it instead of letting the idempotent path insert a duplicate ER.

## Changes (standard 5 integration points for a new import type)

- **`lib/dbservice/utils/import_worker.ex`** — register `student_enrollment` in `processor_map` + add `process_student_enrollment_record/1` (runs the strict guard, then `create_enrollments/2`).
- **`lib/dbservice/services/enrollment_service.ex`** — add `validate_row_memberships_absent/2` and its helpers.
- **`lib/dbservice/constants/mappings.ex`** — register the type on only the enrollment columns: `student_id` / `student_apaar_id` (optional) and `auth_group`, `academic_year`, `grade`, `batch_id`, `school_code`, `start_date` (required). `grade_id` is derived from `grade` automatically.
- **`lib/dbservice/data_import.ex`** — `format_type_name/1` → "Enroll Existing Students".
- **`lib/dbservice_web/controllers/template_controller.ex`** — add to the CSV-template whitelist.
- **`lib/dbservice_web/live/import_live/new.ex`** — add the dropdown option.
- Basic-auth protected route wired through `import_controller.ex` / `router.ex` / `hooks.js`.

## Notes / semantics

- Reused `create_new_group_user/1` writes the ER with `is_current: true` (matching the normal `student` import), so backfilled rows read as *current* enrollments. If some historical rows should instead be closed/non-current, that's a separate follow-up (the CSV has `start_date` but no `end_date`/`is_current` column today).

## Testing

- `mix compile --warnings-as-errors` — clean
- `mix format --check-formatted` on changed files — clean
- `mix credo` on the logic + test files — no issues
- `mix test test/dbservice/services/enrollment_service_test.exs` — the 6 new `validate_row_memberships_absent/2` tests pass. (One unrelated, pre-existing failure in `handle_group_user_enrollment/1 updates existing group user` exists on this branch independent of these changes.)
- Runtime verification (importing a CSV against a test DB and confirming `group_user`/`enrollment_record` rows appear, plus the negative cases) pending manual run via `/imports`.

### DIFF
--- a/lib/dbservice/constants/mappings.ex
+++ b/lib/dbservice/constants/mappings.ex
@@ -89,7 +89,13 @@ defmodule Dbservice.Constants.Mappings do
         "update_incorrect_grade_to_correct_grade",
         "update_incorrect_auth_group_to_correct_auth_group"
       ],
-      optional: ["batch_movement", "student_update", "dropout", "re_enrollment"],
+      optional: [
+        "batch_movement",
+        "student_update",
+        "dropout",
+        "re_enrollment",
+        "student_enrollment"
+      ],
       type: :string
     },
     "student_father_name" => %{
@@ -327,7 +333,8 @@ defmodule Dbservice.Constants.Mappings do
         "update_incorrect_batch_id_to_correct_batch_id",
         "update_incorrect_school_to_correct_school",
         "update_incorrect_grade_to_correct_grade",
-        "update_incorrect_auth_group_to_correct_auth_group"
+        "update_incorrect_auth_group_to_correct_auth_group",
+        "student_enrollment"
       ],
       type: :string
     },
@@ -341,7 +348,7 @@ defmodule Dbservice.Constants.Mappings do
     },
     "auth_group" => %{
       db_field: "auth_group",
-      required: ["student", "teacher_addition", "re_enrollment"],
+      required: ["student", "teacher_addition", "re_enrollment", "student_enrollment"],
       optional: [],
       type: :string
     },
@@ -353,7 +360,8 @@ defmodule Dbservice.Constants.Mappings do
         "teacher_addition",
         "teacher_batch_assignment",
         "dropout",
-        "re_enrollment"
+        "re_enrollment",
+        "student_enrollment"
       ],
       optional: ["school_addition"],
       type: :string
@@ -364,7 +372,8 @@ defmodule Dbservice.Constants.Mappings do
         "student",
         "teacher_addition",
         "update_incorrect_grade_to_correct_grade",
-        "re_enrollment"
+        "re_enrollment",
+        "student_enrollment"
       ],
       optional: [
         "batch_movement",
@@ -383,7 +392,8 @@ defmodule Dbservice.Constants.Mappings do
         "teacher_addition",
         "teacher_batch_assignment",
         "update_incorrect_batch_id_to_correct_batch_id",
-        "re_enrollment"
+        "re_enrollment",
+        "student_enrollment"
       ],
       optional: [],
       type: :string
@@ -395,7 +405,8 @@ defmodule Dbservice.Constants.Mappings do
         "update_incorrect_school_to_correct_school",
         "re_enrollment",
         "school_addition",
-        "school_deletion"
+        "school_deletion",
+        "student_enrollment"
       ],
       optional: [],
       type: :string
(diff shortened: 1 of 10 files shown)


lib/dbservice/constants/mappings.ex
```diff
@@ -364,7 +372,8 @@ defmodule Dbservice.Constants.Mappings do
         "student",
         "teacher_addition",
         "update_incorrect_grade_to_correct_grade",
-        "re_enrollment"
+        "re_enrollment",
+        "student_enrollment"
```
suryabulusu, line 376: grade is required for this import, but an invalid or unknown grade is silently dropped by add_grade_id. 

create_grade_enrollment/2 then returns {:ok, "No grade enrollment needed"}, so the row can report success without creating the required grade enrollment. 

Could we return a row-level error when the supplied grade cannot be resolved? 

Also, values such as 11abc are currently parsed as grade 11

lib/dbservice/utils/import_worker.ex
```diff
@@ -848,6 +849,33 @@ defmodule Dbservice.DataImport.ImportWorker do
...
+    else
+      case Users.get_student_by_id_or_apaar_id(record) do
+        nil ->
+          {:error,
+           "Student not found. student_id: #{record["student_id"]}, apaar_id: #{record["apaar_id"]}"}
+
+        student ->
+          student = Dbservice.Repo.preload(student, [:user])
+
+          case DataImport.StudentEnrollment.create_enrollments(student.user, record) do
```
suryabulusu: Could we make this operation atomic per student? create_enrollments writes auth-group, school, batch and grade sequentially. If a later enrollment fails, the earlier writes remain even though this row reports an error. 

Also, create_new_group_user inserts the enrollment record before creating the group-user, so a failure can leave an orphan ER and undermine the idempotent retry behavior described in this PR. 

Please wrap the complete per-student operation in a transaction / Ecto.Multi and roll back on any error.

suryabulusu (commented):
left two comments

lib/dbservice/constants/mappings.ex
Bahugunajii: 1. It's not necessary that we'll always have grade information with us but still want to get students added in our db, so we can't confirm grade is "required".
2. Most of the sheets contains special characters along with the values eg: 11 {extra space OR some char} and these are not generally cleaned in the sheet itself that's why we have this behaviour

lib/dbservice/utils/import_worker.ex
Bahugunajii: Yes we should but it's not always true that we have these 4 group mappings with us. So any one missing will rollback the entire txn and entries won't get created

suryabulusu (approved):

PR #646, opened 2026-07-23
fix(import): validate duplicate identifiers in student_update import (#641)

### LINKED ISSUE #641: Student Update CSV import bypasses duplicate identifier validation
## Problem

The `student_update` CSV import can overwrite a student with a Student ID, APAAR ID, or PEN number that already belongs to another student. The normal create/update flow validates identifier conflicts, but the import path calls the raw update service directly and skips that validation.

This happened on production in import `1574` (1,265 rows), which recreated 10 duplicate APAAR groups after a cleanup.

## Current path

- `ImportWorker.process_student_update_record/1` finds the student.
- It calls `StudentUpdateService.update_student_with_user_data/2`.
- That calls `Users.update_student_with_user/3` without the shared duplicate-identifier validation.

The new database unique indexes will stop duplicate values from being stored, but the import currently halts on the first constraint error after earlier rows may already have been updated.

## Expected

- Validate Student ID, APAAR ID, and PEN conflicts before updating each row.
- Report the conflicting row and identifier clearly.
- Define and test import behavior so a duplicate does not produce an unclear partial import.
- Add regression coverage for duplicate identifiers in `student_update` imports.

Related rollout PR: #581

Bahugunajii:
Fixes #641.

## Problem
The `student_update` CSV import called the raw update service directly, bypassing the duplicate-identifier validation used by the normal create/update flow. A row could move another student's **APAAR ID** (or **PEN number**) onto a different student — recreating duplicate APAAR groups on prod (import 1574, 1,265 rows). Only the DB unique index caught it, surfacing as a raw `Ecto.Changeset` interpolated into a string and halting the run with an unclear message.

## Fix
- Add **`Users.validate_identifier_conflicts/2`** — a pre-write check that mirrors the student table's global unique constraints and returns a clear, human-readable message (`"APAAR ID '…' already exists for another student"`). Only identifiers present in the row are checked; the student being updated is excluded (re-saving its own identifiers is allowed).
- **`StudentUpdateService.update_student_with_user_data/2`** runs the check before writing, so it covers both the CSV import and `update_student_by_student_id/2`.
- The import framework already prepends the CSV row number, so a conflict is persisted as `%{row: N, error: "APAAR ID '…' already exists for another student"}` and the offending row fails cleanly instead of overwriting another student.

### Import behavior
Kept the existing fail-fast model (a row error halts the run), now with a clear, row-attributed message. Rows before the conflict stay applied; the operator fixes the sheet and re-runs. No change to the global import loop.

### Why `student_id` is not checked here
The merged auth-group-scoping work makes `student_id` unique only **within an auth group** (see `create_or_update_student/1`). A global `student_id` check would falsely reject an update to any student whose id is legitimately reused in another auth group (the update row carries the student's own id). So the guard covers only the genuinely-global identifiers the DB enforces: **`apaar_id`, `pen_number`**. (PEN is not yet an importable CSV column; the guard covers it at the service layer for when it is, and for direct service calls.)

## Tests
- Service-level: rejects moving another student's `apaar_id`/`pen_number`; allows re-saving own identifiers; allows a new non-conflicting identifier.
- Users context: unit tests for `validate_identifier_conflicts/2`, including that `student_id` is **not** flagged.
- End-to-end: first `student_update` worker test — a conflicting APAAR row halts the import with a clear error and leaves the other student untouched (reproduces the prod incident).

All new tests pass; `mix format`, `mix credo`, and `mix compile --warnings-as-errors` are clean.

## ⚠️ Pre-existing `main` issues (NOT introduced by this PR, not fixed here)
1. **Duplicate migration version `20260701120000`** — shared by `add_lms_student_ingestion.exs` and `add_user_session_indexes.exs`. This makes `mix ecto.migrate` fail on a fresh DB, so **CI on this PR will be red until it is resolved on `main`** (e.g. renumber one migration). Independent of this change.
2. **Global `student_id` unique index vs auth-group scoping** — the LMS-ingestion migration creates `student_student_id_unique_not_null` (never dropped in any `up`), while the auth-group-scoping feature expects `student_id` not to be globally unique. As a result, 3 tests in `users_test.exs` ("create_or_update_student/1 auth-group scoping") fail on clean `main`. Independent of this change.

### DIFF
--- a/lib/dbservice/users.ex
+++ b/lib/dbservice/users.ex
@@ -691,6 +691,48 @@ defmodule Dbservice.Users do
     end
   end
 
+  @doc """
+  Validates that the globally-unique identifiers in `params` (`apaar_id`, `pen_number`)
+  do not already belong to another student, mirroring the student table's global unique
+  constraints so a conflict is reported clearly before the write instead of surfacing as
+  a raw constraint error.
+
+  `student_id` is deliberately NOT checked here: it is only unique within an auth group
+  (see `create_or_update_student/1`), so a global check would falsely flag a student
+  whose id is legitimately reused in another auth group. Only the identifiers the
+  database enforces globally are validated.
+
+  Only identifiers that are present and non-empty in `params` are checked, so a partial
+  update touching unrelated fields passes untouched. Accepts string or atom keys. When
+  `existing_student` is given, that student is excluded from the check (re-saving its
+  own identifiers is allowed); pass `nil` when validating a brand new student.
+
+  Returns `:ok`, or `{:error, message}` naming the conflicting identifier.
+  """
+  def validate_identifier_conflicts(params, existing_student \\ nil) do
+    apaar_id = identifier_param(params, :apaar_id)
+    pen_number = identifier_param(params, :pen_number)
+
+    cond do
+      identifier_present?(apaar_id) and
+          duplicate_exists?(:apaar_id, apaar_id, existing_student) ->
+        {:error, "APAAR ID '#{apaar_id}' already exists for another student"}
+
+      identifier_present?(pen_number) and
+          duplicate_exists?(:pen_number, pen_number, existing_student) ->
+        {:error, "PEN Number '#{pen_number}' already exists for another student"}
+
+      true ->
+        :ok
+    end
+  end
+
+  defp identifier_param(params, key) do
+    Map.get(params, key) || Map.get(params, Atom.to_string(key))
+  end
+
+  defp identifier_present?(value), do: not (is_nil(value) or value == "")
+
   # Validates that student_id and apaar_id are not duplicated in other students
   # When updating, validates that the NEW values in params don't conflict with other students
   defp validate_identifiers_not_duplicated(student_id, apaar_id, existing_student) do
@@ -726,6 +768,7 @@ defmodule Dbservice.Users do
       case field do
         :student_id -> from(s in Student, where: s.student_id == ^value)
         :apaar_id -> from(s in Student, where: s.apaar_id == ^value)
+        :pen_number -> from(s in Student, where: s.pen_number == ^value)
       end
 
     query =
--- a/lib/dbservice/services/student_update_service.ex
+++ b/lib/dbservice/services/student_update_service.ex
@@ -49,7 +49,12 @@ defmodule Dbservice.Services.StudentUpdateService do
     # Filter out nil/empty values to only update provided fields
     filtered_params = filter_update_params(params)
 
-    Users.update_student_with_user(student, user, filtered_params)
+    # Reject a row that would move another student's identifier onto this student
+    # before writing, so the import fails with a clear message instead of a raw
+    # unique-constraint error (see issue #641).
+    with :ok <- Users.validate_identifier_conflicts(filtered_params, student) do
+      Users.update_student_with_user(student, user, filtered_params)
+    end
   end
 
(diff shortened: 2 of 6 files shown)


lib/dbservice/users.ex
```diff
@@ -691,6 +691,48 @@ defmodule Dbservice.Users do
...
+
+  Only identifiers that are present and non-empty in `params` are checked, so a partial
+  update touching unrelated fields passes untouched. Accepts string or atom keys. When
+  `existing_student` is given, that student is excluded from the check (re-saving its
+  own identifiers is allowed); pass `nil` when validating a brand new student.
+
+  Returns `:ok`, or `{:error, message}` naming the conflicting identifier.
+  """
+  def validate_identifier_conflicts(params, existing_student \\ nil) do
+    apaar_id = identifier_param(params, :apaar_id)
```
deepansh96, line 780: Blocking: please trim the APAAR ID before checking and saving it. I tested this locally with an existing APAAR ID `222222222222`, then updated another Student using ` 222222222222 ` with spaces. The update succeeded, and both values were stored. This means spaces can bypass the duplicate check. Please trim the value, treat a whitespace-only value as empty, and add a CSV test for this case.

lib/dbservice/users.ex
```diff
@@ -691,6 +691,48 @@ defmodule Dbservice.Users do
...
+    pen_number = identifier_param(params, :pen_number)
```
deepansh96, line 781: Question: the new check handles `pen_number`, but I could not find a PEN column or mapping in the `student_update` CSV. Can this CSV path currently receive a PEN value? Should this PR do one of these two things?

1. Add PEN to the CSV mapping and add an import test for a duplicate PEN.
2. Keep PEN as service-level protection only and clearly say that PEN is outside the CSV change in this PR.

I want to confirm which behavior we expect before merging.

Bahugunajii: Will add PEN to the mapping file

lib/dbservice/users.ex
Bahugunajii: Fixed. apaar_id and pen_number are now trimmed on both sides: the `Student` changeset trims them (whitespace-only → nil), and `validate_identifier_conflicts/2` normalizes the incoming value the same way before comparing — so a padded ` 222222222222 ` maps to the stored `222222222222` and the conflict is caught. Added a CSV worker test (padded APAAR halts the import) plus unit tests for the trim and the whitespace-only case.

lib/dbservice/users.ex
Bahugunajii: Went with option 1. Added `student_pen_number` to the `student`/`student_update` CSV mapping, and added an import test that halts when a row moves another student's PEN onto a different student. Updated the PR description accordingly.

Bahugunajii:
@suryabulusu on both points: (1) trimmed the apaar (and pen) from the sheet — padded values can no longer bypass the duplicate check. (2) You're right the pre-existing issues are fixed on main — I merged latest main into the branch, which resolves the duplicate migration version (`add_user_session_indexes` renamed to `20260724110000`) and drops the global `student_id` unique index/constraint. Updated the description to reflect this. A fresh full-suite run now matches main's baseline exactly (21 pre-existing failures, 0 in the identifier area).

deepansh96 (approved):
Rechecked the latest fixes locally. The focused APAAR/PEN update and import tests pass.

PR #647, opened 2026-07-23
Enforce canonical YYYY-YYYY academic year on enrollment_record

Bahugunajii:
## Summary

Enforces the canonical `YYYY-YYYY` academic year format on `enrollment_record`, rejecting short forms like `"2026-27"` that leaked in through imports.

- **Changeset validation** (`enrollment_record.ex`): `validate_format` on `academic_year` against `^\d{4}-\d{4}$`, plus a matching `check_constraint` so DB-level rejections surface as a friendly changeset error.
- **DB check constraint** (migration): `enrollment_record_academic_year_format`. `NULL` is still allowed because `auth_group` enrollment records legitimately carry no academic year.
- The migration **fails loudly** if any existing rows violate the format, so bad rows are resolved before the constraint is added rather than silently skipped.
- Updated existing tests and CSV import fixtures from the short form to the canonical `YYYY-YYYY` form.

## Notes

- `academic_year IS NULL` is intentionally permitted (auth_group records).
- Producers of `academic_year` (imports, `resolve_academic_year`) pass values through from params/CSV, so this validation is the enforcement point.

## Testing

- `mix format --check-formatted` — clean
- `mix test` on all affected files — 80 tests, 0 failures


lib/dbservice/enrollment_record/enrollment_record.ex
```diff
@@ -48,9 +48,25 @@ defmodule Dbservice.EnrollmentRecords.EnrollmentRecord do
...
     |> validate_dates_of_enrollment
   end
 
+  # Enforces the canonical YYYY-YYYY academic year (e.g. "2026-2027"), rejecting
+  # short forms like "2026-27" that have leaked in via imports. Skips nil, so
+  # auth_group enrollment records (which carry no academic year) are unaffected.
+  # Paired with the DB check constraint of the same name as a hard backstop.
+  defp validate_academic_year_format(changeset) do
+    changeset
+    |> validate_format(:academic_year, ~r/\A\d{4}-\d{4}\z/,
```
suryabulusu, line 70: do we accept years like 2025-2027 .. but does it make sense as part of "academic year"

priv/repo/migrations/20260723120000_add_academic_year_format_check.exs
```diff
@@ -0,0 +1,37 @@
...
+          AND academic_year !~ '^[0-9]{4}-[0-9]{4}$'
+      ) THEN
+        RAISE EXCEPTION 'enrollment_record.academic_year has values not in YYYY-YYYY format; resolve before adding constraint';
+      END IF;
+    END $$;
+    """)
+
+    execute("""
+    ALTER TABLE enrollment_record
+    ADD CONSTRAINT enrollment_record_academic_year_format
```
suryabulusu: claude's comment. i didn't understand it fully

the plain ADD CONSTRAINT ... CHECK takes an ACCESS EXCLUSIVE lock and scans the whole table to validate. 

Simple example: every registration session write on the table blocks behind the validation scan, and everything queued behind it blocks too. 

Fix is the standard two-step: ADD CONSTRAINT ... NOT VALID, then VALIDATE CONSTRAINT (only takes SHARE UPDATE EXCLUSIVE, writes keep flowing).

lib/dbservice/enrollment_record/enrollment_record.ex
```diff
@@ -48,9 +48,25 @@ defmodule Dbservice.EnrollmentRecords.EnrollmentRecord do
       :subject_id
     ])
     |> validate_required(required_fields)
+    |> validate_academic_year_format
```
deepansh96, line 51: Blocking: this validation breaks the existing enrollment-record create and update controller tests. Both tests still submit `some academic_year` / `some updated academic_year`, so they now receive 422 instead of the expected 201/200. I reproduced this on a fresh database: the PR has 5 full-suite failures versus 3 on its base. Please update those request payloads and response assertions to canonical `YYYY-YYYY` values, then rerun the full suite.

lib/dbservice/enrollment_record/enrollment_record.ex
deepansh96, line 70: This checks only the `NNNN-NNNN` shape, so values such as `2025-2027` pass. The local database already has 6 such rows. Is that intended? If “canonical academic year” means consecutive years, please enforce `end_year = start_year + 1` in both the changeset and database constraint, with tests. If shape-only validation is intentional, please state that explicitly in the PR.

deepansh96:
The Swagger schema still documents `academic_year` as `2024-25`, but this PR makes that value invalid. Please update the example to `2024-2025` and check other public examples so API consumers are not told to send a rejected format.

lib/dbservice/enrollment_record/enrollment_record.ex
Bahugunajii: Enforced consecutive years. `validate_consecutive_academic_year` now rejects a well-formed but non-consecutive range like `2025-2027` (end must be start + 1), and the DB CHECK constraint enforces the same (`left(...)::int + 1 = right(...)::int`, CASE-guarded so the casts only run on shape-valid values). Added tests for `2025-2027` / `2027-2025` / `2026-2026`. The migration's fail-loud pre-check also flags non-consecutive rows now — heads up that the ~6 existing non-consecutive rows you mentioned will need cleanup before this deploys (which is the intended fail-loud behavior).

lib/dbservice/enrollment_record/enrollment_record.ex
Bahugunajii: Fixed. The create/update controller tests now send canonical `YYYY-YYYY` values instead of `some academic_year`. One subtlety: the create action is an upsert keyed on (user, group, group_type, academic_year), and `get_ids_create_attrs` builds a fixture whose default is `2022-2023` — so I used distinct values (`2030-2031` create, `2031-2032` update) to keep the create path a 201 rather than matching the fixture row. enrollment_records_test + enrollment_record_controller_test are green (12/12), and a broad run across import/services/student suites is 120/0.

priv/repo/migrations/20260723120000_add_academic_year_format_check.exs
Bahugunajii: Done — switched to the two-step. The migration now runs `ADD CONSTRAINT ... CHECK (...) NOT VALID` (brief lock, no scan) then `VALIDATE CONSTRAINT` (SHARE UPDATE EXCLUSIVE, writes keep flowing). Made it `@disable_ddl_transaction true` so the two run in separate transactions — otherwise the NOT VALID lock would be held through the VALIDATE scan and defeat the purpose.

lib/dbservice/enrollment_record/enrollment_record.ex
Bahugunajii: Good question — you and deepansh both flagged this. Now enforced: an academic year must span consecutive years, so `2025-2027` is rejected (changeset + DB constraint + tests).

Bahugunajii:
Updated the Swagger examples: enrollment_record `academic_year` example `2022` → `2022-2023`, and the student schema description `(e.g., 2024-25)` → `(e.g., 2024-2025)`. Checked the rest — the teacher/student examples were already canonical. (The generated `priv/static/swagger.json` is gitignored and regenerates from these sources.)

Bahugunajii:
Merged latest `main` into the branch (it was ~116 commits behind). Two things this resolved/surfaced:

**Migration-version collision (was blocking a fresh migrate):** the branch still had `add_user_session_indexes` at the stale `20260701120000`, colliding with `add_lms_student_ingestion`; the merge picked up main's rename to `20260724110000`, so a fresh `mix ecto.migrate` runs cleanly again.

**Merge conflict** was in `test/dbservice/services/enrollment_service_test.exs` — main had rewritten it for the exclusive-current unique index (single current enrollment) while this branch had the older multi-current setup. Took main's structure and re-applied the canonical `YYYY-YYYY` values.

**Academic-year canonicalization across merged tests:** main's newer tests (holistic-mentorship, student-eligibility-cleanup, LMS ingestion) created enrollment records with the short form `2026-27` — in both Elixir payloads and raw-SQL inserts — plus one non-consecutive `2026-2028`. Canonicalized all of them (the intentional rejection tests in `enrollment_records_test.exs` are left as short forms on purpose).

**Full-suite validation (fresh DB, `--seed 0`):** the academic-year-directly-affected suites are green (44/44), and there are **0** academic-year format/consecutive/constraint failures across the whole suite. For reference, `origin/main` on a fresh single run is 21 failures and this branch is 26 — all in the same pre-existing environmental buckets (some holistic/LMS tests commit Grade rows outside the sandbox, causing order-dependent `MultipleResultsError`, plus a few holistic domain-constraint failures). None are related to this PR; the small delta is test-ordering shift from the 6 tests this PR adds. Worth a separate look at that test-isolation issue, but it's out of scope here.

deepansh96 (approved):
Rechecked the fixes locally and pushed the small correction to keep the invalid academic-year rejection test valid. The focused academic-year tests pass.

PR #649, opened 2026-07-23
Remove global Student ID uniqueness constraint

deepansh96:
## Summary

- drop `student_student_id_unique_not_null` with a concurrent migration
- remove the matching Ecto `unique_constraint`
- update the schema constraint test to allow duplicate Student IDs while keeping APAAR IDs unique

## Why

Student IDs are scoped to an auth group. The global database index rejects valid edits and creation when another auth group already uses the same Student ID.
Existing auth-group-scoped application matching remains unchanged. APAAR and PEN uniqueness remain global.

## Checks

- `mix check` passes
- targeted student identifier constraint test passes
- full `mix test`: 569 pass, 2 unrelated existing `LmsCurriculumTest` failures caused by duplicate chapter codes


deepansh96:
@codex

PR #656, opened 2026-07-26
Fix duplicate batch/exclusive group_user memberships (NVS)

Bahugunajii:
## Context

Dhyanesh reported two NVS data-integrity defects. Both stem from the same root cause: `group_user` memberships are updated in place / created on batch/enrollment moves but never cleaned up, and nothing enforces the "one current membership per exclusive type" invariant at the DB level.

- **Issue 1** — users with more than one `is_current=true` `enrollment_record` for an exclusive `group_type` (`auth_group`/`school`/`grade`).
- **Issue 2** — one current batch in ER but multiple batch rows in `group_user`. Root cause: `BatchEnrollmentService.update_batch_user/3` repointed only the *first* batch `group_user` (`Enum.find`) and never deleted the rest, so `LmsStudentUpdate.replace_existing_group_user/3` hit `{:error, :multiple_group_users}` and grade/stream edits failed.

## Key finding

`group_user.inserted_at` is **not** a reliable "latest" signal — in-place repointing freezes it, so a stale batch row can be newer than the current one (verified on student `22107508`: stale batch 697 inserted *after* current batch 922). The dedup is therefore **ER-anchored** (keep the membership matching the current enrollment_record; `inserted_at` only breaks ties). This is also why it's safe for legitimately multi-batch students — every current batch is preserved.

## Changes

**Root-cause code fix**
- `update_batch_user/3` reconciles batch `group_user` **scoped to the target batch's program** (collapse to one, delete stale/dups); `update_grade_user/3` collapses all grade rows. New shared `reconcile_group_user/3`.
- `LmsStudentUpdate.replace_existing_group_user/3` **self-heals** duplicates instead of erroring — unblocks grade/stream edits even before the data cleanup.

**Data remediation migrations** (reversible — each snapshots changed rows into `bak_*` tables and restores in `down`)
- `20260725120000` — collapse >1 current exclusive ER (keep newest `start_date`), mirror into `group_user`.
- `20260725120001` — ER-anchored batch `group_user` dedup; users whose current batch has no `group_user` are skipped for manual review (`RAISE NOTICE`).
- `20260725120002` — guarded partial unique index `enrollment_record_current_exclusive_type_unique` on `(user_id, group_type) WHERE is_current AND group_type IN exclusive`. **Batch excluded** (legitimately multi-program).

Dedup SQL uses window-function / anti-join plans — the initial correlated-subquery versions ran 10+ hours on the 1.9M-row `group_user` table.

## Verified on dev
- Exclusive current-ER duplicates: **8,572 → 0**
- Student `22107508`: 2 batch `group_user` → **1** (batch 922, matching the ER)
- True Issue-2 population (1 current batch ER + >1 group_user): **0**; ~5,279 legitimately multi-batch users preserved
- Manual-review skipped: 110 (Issue 1), 13 (Issue 2)
- `mix check` green; `mix test` — 3 failures, all pre-existing on `main` (auth-group `student_id` unique-index tension from #558), unrelated to this PR

## Notes for reviewers / ops
- The exclusive dedup touches ~8,500 users (grade/school far beyond the reported 141) — please confirm those are genuine bug rows before running on prod. `bak_*` tables make it reversible.
- Two `enrollment_service_test.exs` tests were updated: they previously seeded multiple current exclusive ERs directly, which the new invariant forbids. That file overlaps with #647's academic-year formatting; expect a trivial conflict on the two edited tests when both land.

### DIFF
--- a/lib/dbservice/lms_student_update.ex
+++ b/lib/dbservice/lms_student_update.ex
@@ -699,11 +699,27 @@ defmodule Dbservice.LmsStudentUpdate do
     end
   end
 
-  defp replace_existing_group_user(old_group_users, user_id, new_group_id) do
-    case old_group_users do
-      [] -> GroupUsers.create_group_user(%{user_id: user_id, group_id: new_group_id})
-      [old_group_user] -> GroupUsers.update_group_user(old_group_user, %{group_id: new_group_id})
-      _ -> {:error, :multiple_group_users}
+  defp replace_existing_group_user([], user_id, new_group_id),
+    do: GroupUsers.create_group_user(%{user_id: user_id, group_id: new_group_id})
+
+  defp replace_existing_group_user(old_group_users, _user_id, new_group_id) do
+    # Self-heal legacy duplicates instead of erroring: keep one row (preferring one
+    # already pointing at the target), repoint it, and delete the rest. Callers scope
+    # `old_group_users` (batch by program, other types by type), so collapsing the
+    # list to a single row is safe and unblocks grade/stream edits for students who
+    # still carry duplicate batch memberships.
+    {keep, drop} =
+      case Enum.split_with(old_group_users, &(&1.group_id == new_group_id)) do
+        {[match | extra], others} -> {match, extra ++ others}
+        {[], [first | rest]} -> {first, rest}
+      end
+
+    Enum.each(drop, &GroupUsers.delete_group_user/1)
+
+    if keep.group_id == new_group_id do
+      {:ok, keep}
+    else
+      GroupUsers.update_group_user(keep, %{group_id: new_group_id})
     end
   end
 
--- a/lib/dbservice/services/batch_enrollment_service.ex
+++ b/lib/dbservice/services/batch_enrollment_service.ex
@@ -12,6 +12,7 @@ defmodule Dbservice.Services.BatchEnrollmentService do
   alias Dbservice.EnrollmentRecords.EnrollmentRecord
   alias Dbservice.Batches.Batch
   alias Dbservice.Groups.Group
+  alias Dbservice.Groups.GroupUser
   alias Dbservice.Statuses.Status
   alias Dbservice.Grades.Grade
   alias Dbservice.Users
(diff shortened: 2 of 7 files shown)


deepansh96:
When moving a student from one batch to another within a program, this code marks all their current batch enrollment records as inactive, including batches in other programs. However, it keeps the batch membership for those other programs, so the two records no longer match.

Fix: update only the enrollment record for the program being changed and add a test for a student enrolled in two programs.

deepansh96:
This PR adds a unique index that allows only one current enrollment record per student for `school`, `grade`, and `auth_group`, but some application paths are not prepared for this constraint. The enrollment-record changeset does not register the new index, so normal writes can raise an unhandled database exception instead of returning a validation error. The auth-group reactivation and dropout-restoration paths also use bulk updates that bypass the changeset and can conflict with an existing current enrollment or leave partial changes.

**Fixes:**
- Register `enrollment_record_current_exclusive_type_unique` in the Enrollment Record changeset and return a clear 422/domain error for conflicts.
- Update auth-group reactivation to end any other current auth-group enrollment before reactivating the intended one, in a single transaction.
- Update dropout restoration to restore only the exact previous enrollments, reject conflicts with newer current enrollments, and make the operation atomic.
- Add tests for `school`, `grade`, and `auth_group` conflicts, including reactivation, dropout undo, the 422 response, and rollback with no partial changes.

deepansh96:
`update_batch_user/3` and `update_grade_user/3` no longer use the `group_users` argument. Their callers still load all of the student’s memberships and pass the list in, but the functions ignore it and query the database again. This adds an unnecessary query and makes the function interface misleading.

**Fix:** change these functions from `/3` to `/2`, remove the unused `group_users` argument, and remove the caller-side membership queries wherever the list is no longer used.

Bahugunajii:
Addressed all four in 9f5ccae:

1. **Batch move inactivating other programs' ERs** — `handle_batch_enrollment/5` now ends only the current batch ERs in the *target batch's program* (new `end_current_batch_enrollments_in_program/3`), mirroring the program-scoped `update_batch_user/2` reconcile. Added a test: a two-program student moving batch in program A keeps program B's batch ER current.

2 & 3. **Exclusive-current index not wired into app paths** —
   - Registered `enrollment_record_current_exclusive_type_unique` as a `unique_constraint` on the ER changeset, so a conflicting single-record write returns a changeset error instead of raising `Ecto.ConstraintError`.
   - `reactivate_auth_group_enrollment/2` now ends any *other* current auth_group ER first, then revives exactly one row (previously a bulk `update_all` flip that could create two current rows), inside the existing transaction.
   - Dropout undo (`restore_global_enrollments`) now rejects a restore that would collide with a newer current exclusive enrollment (`ensure_no_exclusive_conflict/2`) and returns a clean error — the flow was already transactional. Tests: exclusive-conflict returns a changeset error (+ `batch` is exempt); reactivation leaves exactly one current auth_group ER after ending a stray one.

4. **Unused `group_users` arg** — `update_batch_user`/`update_grade_user` are now `/2`; removed the redundant caller-side membership queries in `student_controller`, `batch_movement`, and `teacher_batch_assignment`, and updated the existing tests to the `/2` arity.

`mix format`, `mix credo`, and the batch/enrollment/users/dropout/data-import suites are green.

deepansh96 (approved):
Approved after reviewing the follow-up changes.

PR #677, opened 2026-08-05
fix(resources): make curriculum_id optional on problem/test-problem endpoints (#651)

### LINKED ISSUE #651: Remove curriculum_id query parameter from GET /api/resource/test/:id/problems
Remove `curriculum_id` from `GET /api/resource/test/:id/problems`, because every problem can have exactly one `resource_curriculum` row, hence filtering is not required on `curriculum_id`.

_Note_: It should not break existing deployed code sending `curriculum_id` query parameter, just skip using it in that case.

Bahugunajii:
## Summary

Fixes #651.

Each problem/topic belongs to exactly one `resource_curriculum` row, so `curriculum_id` filtering is redundant. This makes `curriculum_id` **optional** on the affected read endpoints, while keeping deployed clients that still send it working unchanged (per the issue note, it will be removed later once there are no such clients).

## Endpoints made curriculum_id-optional

1. **`GET /api/resource/test/:id/problems`** — query param `curriculum_id` now optional (also `lang_code`, which already had an all-languages path).
2. **`GET /api/problems?topic_id=…`** — query param `curriculum_id` now optional.
3. **`GET /api/resource/problem/:problem_id/:curriculum_id`** — added a curriculum-less route `GET /api/resource/problem/:problem_id` (same action); the `/:curriculum_id` route is kept for backward compatibility.

## How it stays backward compatible

`curriculum_id` was only ever used to *select* which `resource_curriculum` to surface, and the JSON/query layers already fall back to the single row when there's no explicit match. Concretely:

- **test-problems**: `get_problems_by_test/2` and `get_problems_by_test_and_language/3` default `curriculum_id` to `nil`; a nil `requested_curriculum_id` makes the JSON layer use the problem's single `resource_curriculum` (`List.first`).
- **/api/problems**: the two `fetch_problems/2` clauses are collapsed into one requiring only `topic_id`. The curriculum filter moved into a conditional `maybe_filter_by_curriculum/2` (skipped when nil), and `requested_curriculum_id` is derived from the joined `resource_curriculum` row.
- **/api/resource/problem/:problem_id**: a nil `curriculum_id` falls back to the problem's single `resource_curriculum` row.

For every combination that worked before (`curriculum_id` present, with or without `lang_code`), behavior is identical.

## Testing

- New tests cover, for each endpoint: `curriculum_id` omitted (all-languages **and** single-language), the top-level curriculum fields still resolving from the single row, comprehension paragraph still present, 404 when a problem has no curriculum mapping, and the existing `curriculum_id`-present paths (backward compat).
- `mix test` for the problem/resource/search suites — 33 tests, 0 failures.
- `mix phx.routes` confirms all three problem routes register with no conflicts.
- `mix format --check-formatted` and `mix credo` pass.

### DIFF
--- a/lib/dbservice/resources.ex
+++ b/lib/dbservice/resources.ex
@@ -54,12 +54,14 @@ defmodule Dbservice.Resources do
   ## Parameters
     - test_id: ID of the test resource
     - lang_code: Code of the language to fetch problems in (e.g., "en", "hi")
-    - curriculum_id: ID of the curriculum to get difficulty level
+    - curriculum_id: optional; deprecated. Each problem has exactly one
+      resource_curriculum row, so no curriculum filtering is needed. Accepted
+      for backward compatibility and ignored when nil (issue #651).
 
   ## Returns
     - List of problem resources with their metadata from problem_lang table and difficulty_level from resource_curriculum
   """
-  def get_problems_by_test_and_language(test_id, lang_code, curriculum_id) do
+  def get_problems_by_test_and_language(test_id, lang_code, curriculum_id \\ nil) do
     language = from(l in Language, where: l.code == ^lang_code, select: l) |> Repo.one()
 
     case language do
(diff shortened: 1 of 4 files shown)


ankurcdoshi (changes requested):
Each problem & topic also belongs to single curriculum only, hence need to make `curriculum_id` **optional** from following apis as well. Right now keep it for backward compatibility, but later on we will remove it once there are no clients of existing api.

- `GET /api/problems?topic_id=192&curriculum_id=1`
- `GET /api/resource/problem/3337/1`

Bahugunajii:
Addressed in 0040a44 — extended the same optional-`curriculum_id` treatment to the two endpoints you flagged:

- `GET /api/problems?topic_id=…` — `curriculum_id` query param now optional.
- `GET /api/resource/problem/:problem_id/:curriculum_id` — added `GET /api/resource/problem/:problem_id`; the `/:curriculum_id` route is kept for backward compatibility.

Both fall back to the problem's single `resource_curriculum` row when `curriculum_id` is omitted, and behave exactly as before when it's supplied. Tests added for each. Left them in place (not removed) since existing clients still send it, as you noted.

ankurcdoshi (approved):

